PR #7, opened 2016-03-04
Use contain-rs' maps

gereeter:
Replacing constructions like `Vec<(K, V)>` and `Vec<Option<V>>` with the equivalent maps provided by @contain-rs makes it more obvious how those structures are being used, allows for a much broader set of operations, and doesn't require in-tree maintenance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ time = "0.1.34"
 fnv="1.0.2"
 rc="0.1.1"
 linear-map = "0.0.4"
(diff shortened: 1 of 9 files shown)


frankmcsherry:
Thanks! I'll need a day or so to check this out. There is a deadline tomorrow for some other work. But, it seems like a good change. :)

frankmcsherry:
This looks great, thanks! As always, it is much appreciated! :D

PR #8, opened 2016-06-10
Fix type annotation in consolidate

utaal:
Encountered this while reading code.


utaal:
This is actually not relevant anymore. Closing.

Issue #6, opened 2016-02-24
Query2 !


kali:
Hey ! Still obsessing on query2 :)

```rust
            let mut input = root.scoped::<u64, _, _>(move |builder| {
                let (input, stream) = builder.new_input::<((K, _), i32)>();
                let collections = Collection::new(stream);

                let group:Collection<_, (K,SaneF32)> = collections.group(|_,vs,o| {
                    let v:f32 = vs.map(|(sane,weight):(&SaneF32,i32)| sane.0*weight as f32).sum();
                    o.push( (SaneF32(v), 1) );
                });
                let count:Collection<_, (bool, u32)> = group.map(|(_,_):(K,SaneF32)| (true, 1)).consolidate();
                let count:Collection<_, (bool, i32)> = count.group(|_,vs,o| {
                    let c:i32 = vs.map(|(c,weight):(&u32,i32)| *c as i32 * weight).sum();
                    o.push((c, 1));
                });
                count.inspect(move |rec| result.store((rec.0).1 as usize, Relaxed));

            input
            });

            for (_i,chunk) in bibi.enumerate() {
                for item in chunk {
                    input.send(((item.0, SaneF32(item.1)), 1));
                }
            //    input.advance_to(i as u64 + 1);
            //    root.step();
            }

```

This is what I tried so far (with variation on the way to push the input as suggested byt the commented bit). Happy to say it is produces the correct result.

The problem I have is, it makes a gigantic memory consumption, compared to the timely impl, so it swaps and is quite slow.

diff-d:

![tdd](https://cloud.githubusercontent.com/assets/53657/13281890/a26086be-dae6-11e5-993f-70c9a6a97b0e.png)

timely-d:

![timely](https://cloud.githubusercontent.com/assets/53657/13281916/bebcfb58-dae6-11e5-97a1-601c3cf9b860.png)

This is for query 2A (X=8). Timely and the hand made impl made it under 500MB, differential activates 25GB...

Is it something that is expected ? Can i mitigate it somehow ?

frankmcsherry:
This is expected (and, apologies for the surprise). What differential dataflow needs help with is understanding combiners. The aggregation-in-place that timely does is hard for differential, because it thinks of each of the `(key, val)` pairs as different, even though you would like them just to be summed up by `key`, and forget that they had different `val`s in the first place.

[...]

PR #18, opened 2017-03-18
Support for general rings instead of just counts

frankmcsherry:
Differential dataflow has previously used `isize` or `i32` for the counts associated with updates. This pull request generalizes this to mathematical rings, which are types supporting addition, subtraction, and multiplication. 

This flexibility allows the user to use `isize` or `i32` as they see fit, but also allows us additional flexibility to accumulate more general quantities. For example, #6 suffers from the inability of differential dataflow to accumulate inputs in-place, maintaining records to sum as different updates instead of accumulating them. We can now write (e.g. in `examples/ring.rs`):

```rust
    let (input, data) = scope.new_input::<((usize, isize), isize)>();

    // move `val` into the ring component.
    data.map(|((key,val),d)| (key, Default::default(), RingPair::new(val,d)))
	.as_collection()
        .consolidate();
```

which allows us to move accumulable data into the ring, where differential dataflow is now able to accumulate the contributions.

Each ring element has an additive identity ("zero"), and accumulated updates can be discarded when their ring element equals this zero. They can not otherwise be discarded, even though individual components may be zero.

Restrictions include the fact that some common types are not rings, for example `f32` and `f64`. Also, the simplest implementation of rings as tuples of other rings does not work, because tuples do not implement `Add`, `Sub`, `Neg`, `Mul`, etc. At the moment we provide a `RingPair` type which acts as a re-implementation of pair supporting `Ring`, but it is worth thinking about whether this is the best we can do.

As part of this PR, trace implementations should (or at least, `rhh_k` does) compact updates as they are accumulated. The in-place updating has no benefits if we just append them all, of course. This can result in a performance hit if the compaction happens when it used to not happen, and may provide no benefit. The point at which compaction happens should be parameterizable, but it is currently just a constant in `rhh_k.rs`. There are some further thoughts on how the data should be store (e.g. as traces, which are tight representations, or as buffers `Vec<(Data, Time, Ring)>` which are easily re-used as the data are manipulated.

There is a new example, `examples/ring.rs`, which demonstrate in-place updating and ring functionality by introducing an unbounded number (`usize::max_value()`) of records with some bounded number of (key, val) pairs with some bounded number of keys (and junk vals to accumulate). It runs and reports the rate at which it consumes inputs, for example:

```
Echidnatron% cargo run --release --example ring -- 1000 10000000
   Compiling differential-dataflow v0.0.4 (file:///Users/mcsherry/Projects/differential-dataflow)
    Finished release [optimized + debuginfo] target(s) in 25.79 secs
     Running `target/release/examples/ring 1000 10000000`
tuples: 10000000,	elts/sec: 5761822.478278253
tuples: 20000000,	elts/sec: 5914238.979648239
tuples: 30000000,	elts/sec: 5828678.981620855
tuples: 40000000,	elts/sec: 5991998.556864298
...
tuples: 990000000,	elts/sec: 7035803.761681548
tuples: 1000000000,	elts/sec: 7033942.504046317
tuples: 1010000000,	elts/sec: 7034213.746759261
...
```

which gives us a sense for the cost of in-place compaction. The benefit is that without it, one billion of these records would have occupied 12GB+, and here it wobbles around 150MB.

The rate without compaction is initially something like 20M elts/sec, and the cost seems to be closely connected to the cost of hashing and radix sorting. If the `hashed()` method (the default hash function) is reduced from 64 to 32 bits, the rate increases to 10M elts/sec, due to fewer radix passes meaning fewer hash invocations. We should experiment with types that have cheaper hash functions to see if it is the passes (and data movement) or the hash evaluation (which we can fix).

### DIFF
--- a/examples/bfs.rs
+++ b/examples/bfs.rs
@@ -40,7 +40,7 @@ fn main() {
             }
 
             let probe = result.map(|(_,l)| l)
-                              .consolidate_by(|&x| x)
+                              .consolidate()
                               .inspect(|x| println!("\t{:?}", x))
                               .probe();
 
--- a/examples/cc.rs
+++ b/examples/cc.rs
@@ -52,7 +52,7 @@ where G::Timestamp: Lattice+Hash+Ord {
                         let min = std::cmp::min(pair.0, pair.1);
                         *pair = (min, min);
                      })
-                     .consolidate_by(|x| x.0);
+                     .consolidate();
 
     // each edge should exist in both directions.
     let edges = edges.map_in_place(|x| mem::swap(&mut x.0, &mut x.1))
@@ -66,7 +66,6 @@ where G::Timestamp: Lattice+Hash+Ord {
 
             inner.join_map(&edges, |_k,l,d| (*d,*l))
                  .concat(&nodes)
-                 // .consolidate()
                  .group(|_, s, t| { t.push((s[0].0, 1)); } )
          })
 }
--- a/examples/dataflow.rs
+++ b/examples/dataflow.rs
@@ -68,7 +68,7 @@ fn main() {
                 query_topics = query_topics.filter(|_| false);
             }
 
-            let probe = query_topics.consolidate_by(|&(_,q,_)| q)
+            let probe = query_topics.consolidate()
                                     .inspect(|&((l,q,t),_,w)| println!("\t(query: {},\tlabel: {},\ttopic:{}\t(weight: {})", q, l, t, w))
                                     .probe();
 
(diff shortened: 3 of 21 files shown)


frankmcsherry:
Fixes #6 and #18.

frankmcsherry:
Note: it would be pretty easy to add support for `f32` and `f64`, and must be done in `ring.rs` if so, but we risk losing determinism of the computation (because float addition is not associative). We would also either (i) need to put a threshold in `is_zero()` or risk unbounded accumulation of "effectively zero" updates that would not cancel even when their inputs were retracted.

Issue #19, opened 2017-03-18
High-resolution updates `group` perf harms `degrees.rs` example.


frankmcsherry:
The `degrees.rs` example computes and maintains the distribution of degrees in a continually changing random graph. It looks roughly like so:

```rust
let degrs = edges.map(|(src, _)| src)
                 .count();

let distr = degrs.map(|(_, cnt)| cnt)
    		 .count();
```

The current `group` implementation has quadratic behavior with respect to the number of distinct times per key, within each batch. With sufficiently large graphs, random changes are unlikely to touch many nodes within the same batch, but they are quite likely to touch the same *counts*, in the second `count` operator.

Here are some timings with the current code, for batch size one:

```
Echidnatron% cargo run --release --example degrees -- 10000 50000 1
worker 0, round 1 finished after Duration { secs: 0, nanos: 71173 }
worker 0, round 2 finished after Duration { secs: 0, nanos: 73827 }
worker 0, round 3 finished after Duration { secs: 0, nanos: 57993 }
...
```

We would hope that with batching the throughput would improve, and it does for a bit. Here are batch sizes ten,
```
Echidnatron% cargo run --release --example degrees -- 10000 50000 10
worker 0, round 10 finished after Duration { secs: 0, nanos: 336847 }
worker 0, round 20 finished after Duration { secs: 0, nanos: 243068 }
worker 0, round 30 finished after Duration { secs: 0, nanos: 238114 }
...
```
 one hundred,
```
Echidnatron% cargo run --release --example degrees -- 10000 50000 100
worker 0, round 100 finished after Duration { secs: 0, nanos: 1528624 }
worker 0, round 200 finished after Duration { secs: 0, nanos: 2137224 }
worker 0, round 300 finished after Duration { secs: 0, nanos: 2024237 }
...
```
and one thousand,
```
Echidnatron% cargo run --release --example degrees -- 10000 50000 1000
worker 0, round 1000 finished after Duration { secs: 0, nanos: 90445490 }
worker 0, round 2000 finished after Duration { secs: 0, nanos: 127967149 }
worker 0, round 3000 finished after Duration { secs: 0, nanos: 128467679 }
```

The average time to process each update in the first batch, for example, goes from 71us to 33us to 15us to 90us as we increase the batch size. That was looking pretty good, until we hit batch size one thousand. The quadratic horror is showing up here, and is probably also tainting the batch size one hundred execution as well.

[...]

PR #21, opened 2017-03-25
Group hires

frankmcsherry:
This pull request substantially changes the implementation of `group` in an attempt to perform closer to "linear" in the number of input updates.

The problem observed is that when a batch of updates contains many distinct times for the same key, the logic of (i) computing all synthetic interesting times and (ii) forming the input and output at these times both took time (at least) quadratic in the size of inputs. Some times quadratic is the right answer (rarely in our examples, but it is possible), but in many cases (notably arbitrarily large rounds of updates) the output is linear in the input, and our computation didn't reflect this. This meant that batching input updates resulted in performance *degradation* rather than improvement, opposite of what is desired.

The underlying `group` implementation is changed to perform tasks (i) and (ii) in time that is meant to be linear (plus sorting) in the inputs multiplied by the number of times that are not `lt` their immediate successor in the total order. There is no guarantee that there will be few such times; in a total order there are none, but in the `(round, iteration)` partial order there may be many. In future work, we could more aggressively re-order the updates, if we see this is a problem.

Running times for the `degrees.rs` example, which determines the degrees of a changing random graph and counts the number of nodes with each degree (this last count suffering many changes to the same few keys) now has performance linear in the batch size:

```
worker 0, round 1 finished after Duration { secs: 0, nanos: 90736 }
worker 0, round 10 finished after Duration { secs: 0, nanos: 215832 }
worker 0, round 100 finished after Duration { secs: 0, nanos: 842466 }
worker 0, round 1000 finished after Duration { secs: 0, nanos: 6758881 }
worker 0, round 10000 finished after Duration { secs: 0, nanos: 55910134 }
worker 0, round 100000 finished after Duration { secs: 0, nanos: 500888218 }
worker 0, round 1000000 finished after Duration { secs: 5, nanos: 405282347 }
```

The `bfs.rs` example has substantially improved performance for the small graph reference (1,000 nodes and 2,000 edges), though it still exhibits degrading performance with larger batches, suggesting that for more complex partial orders there is still work to do (though, for general partial orders it isn't clear how much opportunity we really have).

The main interesting changes lie in the `Interestinator` and `Accumulator` structs. Respectively, these two maintain buffers and logic to (i) determine synthetic interesting times, and (ii) maintain and update accumulations of collections. Each implementation intends to exploit the "narrowness" of the partial order, in that there are decompositions into relatively few disjoint chains, and the frontiers (antichains) at any point cannot be too large.

Because the `group` implementation is now sufficiently exotic, there is also a new `tests/bfs.rs` which applies breadth-first distance labeling to a randomly changing random graph, both using differential dataflow and using a very naive sequential implementation, and checks that they produce identical outputs. This test is included in `cargo test` and should error out if we break the `group` implementation for these cases.

### DIFF
--- a/examples/degrees.rs
+++ b/examples/degrees.rs
@@ -23,13 +23,12 @@ fn main() {
     let nodes: u32 = std::env::args().nth(1).unwrap().parse().unwrap();
     let edges: usize = std::env::args().nth(2).unwrap().parse().unwrap();
     let batch: usize = std::env::args().nth(3).unwrap().parse().unwrap();
-    let k: isize = std::env::args().nth(4).unwrap().parse().unwrap();
 
     let kc1 = std::env::args().find(|x| x == "kcore1").is_some();
     let kc2 = std::env::args().find(|x| x == "kcore2").is_some();
 
     // define a new computational scope, in which to run BFS
-    timely::execute_from_args(std::env::args().skip(5), move |computation| {
+    timely::execute_from_args(std::env::args().skip(4), move |computation| {
 
     	let index = computation.index();
     	let peers = computation.peers();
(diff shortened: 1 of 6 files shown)


frankmcsherry:
Fixes #19.

Issue #20, opened 2017-03-19
Many batch descriptions have empty frontiers


frankmcsherry:
At many points in the code the descriptions assigned to batches (indicating lower and upper frontiers for the times of updates) are simply (and incorrectly) set to `&[]`. This was because I was lazy and a fundamentally bad person. There should be none of these, except possibly for the very last batch as the collection is sealed (as there are no future times).

We could make the `Description` constructor test for non-emptiness and assert otherwise. Until this is cleared up, the descriptions cannot be relied on for neat and interesting optimizations.

frankmcsherry:
Fixed by #22.

PR #22, opened 2017-03-25
Descriptions

frankmcsherry:
This pull request corrects some (all?) of the bounds in batch descriptions, and introduces a few sanity checks to try and ensure they are used correctly.

To recap, each batch of updates track lower and upper frontiers indicating the times contained in the batch. Specifically, the batch contains any and all times that are greater or equal to some element of the lower frontier but not greater or equal to any element of the upper frontier. In essence, these are the times that are "finalized" as the frontier moves from lower to upper.

The `Description` type now asserts that each `lower` should have non-zero length, as anything else would imply a structurally empty batch. The corresponding assertion for `upper` does not need to hold for the last batch, once the frontier closes.

The two implementations, `rhh.rs` and `rhh_k.rs`, both have instrumented their `merge` methods for merging batches with assertion that ensure that two merged batches have the upper frontier of the first not greater than the lower frontier of the second. It is possible to have gaps in these frontiers when there are empty regions of time; we could insist on filling these but it can be challenging given that they exist in part by the absence of information.

### DIFF
--- a/src/operators/group.rs
+++ b/src/operators/group.rs
@@ -227,6 +227,12 @@ where
             // stolen some code from there, so if either looks wrong, make sure to check the other
             // as well.
 
+            // lower bound for use in descriptions.
+            let mut lower = Vec::new();
+            for cap in &capabilities {
+                lower.push(cap.time());
+            }
+
             for index in 0 .. capabilities.len() {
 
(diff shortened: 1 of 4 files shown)


frankmcsherry:
Fixes #20.

PR #51, opened 2017-05-04
Additional tests for Trace.import

utaal:
~~(This is still WIP) I'm not sure the second test should actually pass.~~

Signed-off-by: Andrea Lattuada <andreal@inf.ethz.ch>


frankmcsherry:
Convo with Andrea: 

Because the `trace` handle is not advanced in any way (neither `advance_by` nor `distinguish_since`)  the imported trace should be *exactly equal* to the source trace, and so the second test should pass (and continue to pass).

PR #59, opened 2017-05-10
[wip] first attempt at trace test

utaal:
Here's some code I wrote to figure out `Trace`, `Batcher`, `Cursor`, and co. Not entirely sure it's a useful test (yet).


frankmcsherry:
Looks good. I'm guessing you separated keys and vals out from times because `map_times` wouldn't let you borrow them inside it? We can fix that if it helps (or, pre-cloning them works too); perhaps having a `Vec<(Key, Val, Time, Diff)>` could be more appealing, but no urgency if it is for testing (no need to re-write tests).

PR #62, opened 2017-05-22
examples/bfs: flush graph input after advancing to time=1

vasia:
I've added a missing `.flush()` in the bfs example.


frankmcsherry:
Ahaha. I spent the flight back to the US finding this bug in dijkstra, which was c/p'd from nfs. I hope.it didn't take too much of your time.

PR #64, opened 2017-05-31
No `Rc`s in layers, fixes #61

### LINKED ISSUE #61: Batches contain more `Rc` that one might want
The batch implementations currently use `Rc` somewhat liberally, mostly in support of easily written cursors that can own layers of the batch (the allocations containing keys, values, times, etc). This seems relatively harmless in normal function, as the batches themselves are wrapped with `Rc` and liberally shared where possible; they are intended to be immutable, and treated as such.

However, certain bits of logic would be much simpler if each batch was uniquely owned. The main one I'm thinking of at the moment is durability, where we would like to be able to serialize and deserialize such an object with something like Abomonation, i.e. just mapping some memory, but the existence of `Rc` internal to the type makes this not possible (`Rc` is not safe to abomonate, due to the `Cell` wrappers around the counts, which are meant to keep allocations alive but wouldn't).

Another less critical example, `advance_mut` is the method that transforms a batch into one where times have been advanced to the current frontier, and it is intended to allow this to happen in place if there is unique ownership of the underlying resources. This is relatively easy to sort out if there is one `Rc` around the batch, and more complicated if there are several `Rc`s inside as well.

The main benefit of the internal `Rc` types is that cursors can be written independently for each layer, taking (shared) ownership of the layer rather than worrying about borrowing it. This allows us to write cursors modularly, and compose them easily. If we removed the option of shared ownership, we would seemingly need to either (i) rewrite cursors as acting on borrowed slices, or (ii) write a cursor monolithically, taking (shared) ownership of the whole batch.

Some thoughts on each:

1. Borrowing is annoying, but there is reason to think that we might eventually want this anyhow, if we end up with indices in each trace (from key to locations in a subset of batches). These indices need to be borrowed, rather than shared, (the benefit is that they are updated in place), and so we end up with a borrow anyhow.

2. Monolithic cursor writing doesn't need to be complicated, as the cursor side of the trace layers is pretty simple (compared to the merge logic, which is a bit more complicated and should stay modular, in my opinion). It is probably worth prototyping this out to see if it is horrid or not.

In either case, we may need to rethink "what is a batch" to understand whether it is something pre-`Rc`d, or whether it is something we should expect to wrap in a `Rc`. I am inclined toward making `Batch : Clone` with the understanding that the clone implementation should be relatively efficient (i.e. will be called regularly). If this means an internal `Rc` great, and outside of the batch we won't wrap it in one.

utaal:
Fixes #61.

This removes `Rc`s between Batch layers so that a Batch can be `exhume`d correctly (with Abomonation). Cursors for the layers hold an owning reference to the enclosing full Batch using `OwningRef<Rc<Batch>, _>`.

### DIFF
--- a/src/trace/layers/hashed.rs
+++ b/src/trace/layers/hashed.rs
@@ -7,6 +7,8 @@ use std::rc::Rc;
 use std::default::Default;
 // use std::hash::Hasher;
 
+use owning_ref::OwningRef;
+
 use timely_sort::Unsigned;
 
 use ::hashable::{Hashable, HashOrdered};
@@ -40,7 +42,7 @@ const BLOAT_FACTOR : f64 = 1.1;
 #[derive(Debug)]
 pub struct HashedLayer<K: HashOrdered, L> {
 	/// Keys and offsets for the keys.
-	pub keys: Rc<Vec<Entry<K>>>,	// track upper and lower bounds, because trickery is hard.
+	pub keys: Vec<Entry<K>>,	// track upper and lower bounds, because trickery is hard.
 	/// A lower layer containing ranges of values.
 	pub vals: L,
 }
@@ -51,15 +53,15 @@ impl<K: HashOrdered, L> HashedLayer<K, L> {
 	fn upper(&self, index: usize) -> usize { self.keys[index].get_upper() }
 }
 
-impl<K: Clone+HashOrdered+Default, L: Trie> Trie for HashedLayer<K, L> {
+impl<B, K: Clone+HashOrdered+Default, L: Trie<B>> Trie<B> for HashedLayer<K, L> {
 	type Item = (K, L::Item);
-	type Cursor = HashedCursor<K, L::Cursor>;
-	type MergeBuilder = HashedBuilder<K, L::MergeBuilder>;
-	type TupleBuilder = HashedBuilder<K, L::TupleBuilder>;
+	type Cursor = HashedCursor<B, K, L::Cursor>;
+	type MergeBuilder = HashedBuilder<B, K, L::MergeBuilder>;
+	type TupleBuilder = HashedBuilder<B, K, L::TupleBuilder>;
 
 	fn keys(&self) -> usize { self.keys.len() }
 	fn tuples(&self) -> usize { self.vals.tuples() }
-	fn cursor_from(&self, lower: usize, upper: usize) -> Self::Cursor {
+	fn cursor_from(&self, owned_self: OwningRef<Rc<B>, Self>, lower: usize, upper: usize) -> Self::Cursor {
 
 		if lower < upper {
 
@@ -78,8 +80,8 @@ impl<K: Clone+HashOrdered+Default, L: Trie> Trie for HashedLayer<K, L> {
 				shift: shift,
 				bounds: (lower, upper),
 				pos: pos,
-				keys: self.keys.clone(),
-				child: self.vals.cursor_from(self.keys[pos].get_lower(), self.keys[pos].get_upper())
+				keys: owned_self.clone().map(|x| &x.keys),
+				child: self.vals.cursor_from(owned_self.map(|x| &x.vals), self.keys[pos].get_lower(), self.keys[pos].get_upper())
 			}
 		}
 		else {
@@ -87,8 +89,8 @@ impl<K: Clone+HashOrdered+Default, L: Trie> Trie for HashedLayer<K, L> {
 				shift: 0,
 				bounds: (0, 0),
 				pos: 0,
-				keys: self.keys.clone(),
-				child: self.vals.cursor_from(0, 0),
+				keys: owned_self.clone().map(|x| &x.keys), // &self.keys,
+				child: self.vals.cursor_from(owned_self.map(|x| &x.vals), 0, 0),
 			}
 		}
 	}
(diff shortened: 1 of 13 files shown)


frankmcsherry:
Reading through this, it looks like there might be a large number of places where we could make things generic in a type `T: Deref<Target=L>` for some layer type `L`. Although `OwningRef` is a great example of such an implementor, I bet most of the types (e.g. cursor types) don't need to know about it.

Also, it is broken at the moment, but seems like it could be fixed. Should I take a stab at that, or do you have work in progress?

src/trace/layers/hashed.rs
```diff
@@ -398,17 +403,17 @@ impl<K: HashOrdered+Clone+Default, L: MergeBuilder> HashedBuilder<K, L> {
...
 /// A cursor with a child cursor that is updated as we move.
 #[derive(Debug)]
-pub struct HashedCursor<K: HashOrdered, L: Cursor> {
+pub struct HashedCursor<B, K: HashOrdered, L: Cursor> {
 	shift: usize,			// amount by which to shift hashes.
 	bounds: (usize, usize),	// bounds of slice of self.keys.
 	pos: usize,				// <-- current cursor position.
 
-	keys: Rc<Vec<Entry<K>>>,
+	keys: OwningRef<Rc<B>, Vec<Entry<K>>>,
```
frankmcsherry: This seems like it could possibly be `keys: B` with the constraint `B: Deref<Target=[Entry<K>]>`, which seems appealing in that it hides the details of how we've chosen to wrap up the owned reference to `[Entry<K>]`.

frankmcsherry: Probably multiplies the compile time by five, though. =/

frankmcsherry: Or, maybe this is a horrible mess if it creates a new generic parameter that must be everywhere present throughout lots and lots of types. Let it not be thought of as a uniformly great idea, but might want to think it out.

src/trace/layers/mod.rs
```diff
@@ -10,30 +10,33 @@ pub mod hashed;
...
-	type TupleBuilder: TupleBuilder<Trie=Self, Item=Self::Item>;
+	type TupleBuilder: TupleBuilder<B, Trie=Self, Item=Self::Item>;
 
 	/// The number of distinct keys, as distinct from the total number of tuples.
 	fn keys(&self) -> usize;
 	/// The total number of tuples in the collection.
 	fn tuples(&self) -> usize;
 	/// Returns a cursor capable of navigating the collection.
-	fn cursor(&self) -> Self::Cursor { self.cursor_from(0, self.keys()) }
+	fn cursor(&self, owned_self: OwningRef<Rc<B>, Self>) -> Self::Cursor { self.cursor_from(owned_self, 0, self.keys()) }
```
frankmcsherry: Is there something that prevents this from being a `OwningRef<Rc<Self>, Self>`? I'm trying to track down the place where the need to have the parameter `B` comes in, and this is where I got. :)

frankmcsherry:
I'm thinking a bit, and drinking too so careful, but it seems like we could do a lot of what `OwningRef` does with the trait constraint `B: Deref<Target=T>+Clone` which happens to be implemented by `Rc<T>`. The cursors may need to be generic it types that clone and dereference to their layer type, but .. could work.

This would be to avoid taking a dependence on `OwningRef`, not because that is bad but just to understanding what property we actually need to have (and it seems to be: owned things that can be dereferenced, and construct similar owned things for child fields). Perhaps each layer could indicate a type that does this for their children, as otherwise the type would probably need to have closures in it.

src/trace/layers/mod.rs
utaal: I'm attacking this one comment at a time (it's a bit larger than what fits in my brain at once).

So: It seems like this is the case; I'm trying to get rid of <B>, which may reduce the footprint of the change significantly.

utaal: Ah, no, not the case: `Trie` is implemented for layers, and the `B` in `Rc` is the type of the enclosing `Batch`, which corresponds to `Self` at the top layer, but it doesn't in inner layers.

utaal:
I fixed the test. I'll look into changing `OwningRef` into `B: Deref<Target=T>+Clone`.

src/trace/layers/ordered.rs
```diff
@@ -38,36 +39,37 @@ impl<K: Ord+Clone, L: Trie> Trie for OrderedLayer<K, L> {
 			let child_lower = self.offs[lower];
 			let child_upper = self.offs[lower + 1];
 			OrderedCursor {
-				keys: self.keys.clone(),
-				offs: self.offs.clone(),
+				keys: owned_self.clone().map(|x| &x.keys),
+				offs: owned_self.clone().map(|x| &x.offs),
```
utaal: @frankmcsherry Regarding the use of `Deref<Target=Self>` the problem seems to be that I wouldn't know how to express this: our constraint is that the "thing that dereferences to Self" also supports grabbing another reference that's reachable from `Self` (in this case what's expressed by the closure).

frankmcsherry: That makes sense! After posting the questions, I hopped on to #rust and asked if there was a type that takes an arbitrary `T: Deref` into another `S: Deref<Target=<T as Deref>::Target::...>`, and (i) got no answer, but (ii) realized that this would probably require a closure in the resulting type signature, which puts it off limits for now.

utaal: Also, getting to `OwningRef` was a journey through (arguably) desirable but not-yet-implemented rust type systems features, namely higher-kinded-types or associated type constructors, which could have enabled just using lifetimes and references without most of the pain.

utaal: Relevant RFC (hopefully merged soon): https://github.com/rust-lang/rfcs/pull/1598

frankmcsherry:
Looks good to me, and we can prototype changes as further PRs, if they seem appropriate!

PR #67, opened 2017-07-21
Address name clash introduced by nightly

frankmcsherry:
Rust nightly has introduced new methods `Ord::min()` and `Ord::max()`, which in the noble goal of removing papercuts just migrated them to other people. This PR changes `Lattice::min()` and `Lattice::max()` to be `Lattice::minimum()` and `Lattice::maximum()` instead. 

They may change again, if the Rust folks randomly change things again. This is all nightly Rust, but since you can't load/use shared libraries outside of nightly it is important to have a differential dataflow that builds with nightly.

### DIFF
--- a/src/lattice.rs
+++ b/src/lattice.rs
@@ -16,10 +16,10 @@ pub trait Lattice : PartialOrder {
     /// ```
     /// use differential_dataflow::lattice::Lattice;
     ///
-    /// let min = <usize as Lattice>::min();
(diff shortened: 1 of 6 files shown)


frankmcsherry:
NB: This also swaps the order of some instructions in `join`, as this was a work in progress that it was a pain to reverse out. It's minor (pretty sure), but helpful in the upcoming `advance_mut()` PR.

PR #79, opened 2017-09-08
Merge batcher (work in progress)

frankmcsherry:
This PR is a work in progress, studying the possibility of replacing the `RadixBatcher` with a mergesort based `MergeBatcher`.

The `Batcher` types are ones that accept lumps of `Vec<(D, T, R)>` data, and are asked to "seal" batches up through some `upper: &[T]`. This is where the data are sorted and canonicalized and generally work is done to organize the representation of previously unordered data.

The `RadixBatcher` implementation uses radix sort, which (i) has a pretty decent throughput and (ii) leaves data in their `Vec<(D, T, R)>` form (rather than reallocating). On the downside, it (i) requires `D` elements provide a hash method and their `Ord` implementation respect it, (ii) it does most of its work in a large batch in `seal` making "spiky" work, and consequently (iii) it cannot collapse cancellations in place without additional mechanisms. Also, (iv) we need to post-process the radix sorted data with a sort over records with the same hash value, in case of key collision and also to order by value and time; for whatever reason `sort_unstable` is not great at this.

The `MergeBatcher` by comparison simplifies a bunch of these problems, at the cost of not being as awesome as radix sort. I'm going to enumerate some of the benefits, to think out whether it might be a good idea to switch over generally.

1. Merge sorting is generally simpler than radix sorting. We don't need to have hash values that haven't been tainted by partitioning, the hashing doesn't need to be repeatedly invoked (or cached) to ensure that we maintain the same order, and we don't need to do any finishing sorting to make sure we account for hash collision or multiple values and times with a key.

2. Merge sorting does its work as we receive data, which means that we have the opportunity to compact in place, maintaining a bounded footprint with respect to the compacted representation. This also means we are less likely to have big spikes of work, as the bulk of the `n log n` work happens on insertion, with `n` work at the moment of `seal`. It is also pretty easy to amortize the work of merging, if we would like a very smooth profile.

3. It feels like there is a quantum reduction in complexity, with for example the ability to remove the `_u` variants of operators (which were problematic in a few ways), and the removal of `HashOrdered` wrappers that frustrated the type system (e.g. that `Vec<(OrdWrapper<K>, ..>` is different from `Vec<(K, ..)>` and so required a copy).

The main downside that I foresee is that we could experience performance regressions in cases where the high throughput radix merging really helped. At the moment, on the problems I'm looking at, this is not currently a problem. It *was* a problem, but then some commits repaired that, and times now seem (marginally) better than their `RadixBatcher` equivalents. I'm going to try and report a bit on what I see performance-wise comparing the two, and I'll drop them in as comments here.

### DIFF
--- a/examples/arrange.rs
+++ b/examples/arrange.rs
@@ -97,7 +97,7 @@ fn main() {
             })
             .probe_with(&mut probe)
             .as_collection()
-            .arrange_by_key_u()
+            .arrange_by_key()
             .trace
         });
 
@@ -121,10 +121,10 @@ fn main() {
                 let edges = edges.enter(&dists.scope());
                 let roots = roots.enter(&dists.scope());
 
-                dists.arrange_by_key_u()
+                dists.arrange_by_key()
                      .join_core(&edges, |_k,l,d| Some((*d, l+1)))
                      .concat(&roots)
-                     .group_u(|_, s, t| t.push((*s[0].0, 1)))
+                     .group(|_, s, t| t.push((*s[0].0, 1)))
             })
             .map(|(_node, dist)| dist)
             .consolidate()
@@ -140,11 +140,11 @@ fn main() {
             let (input, query) = scope.new_collection();
 
             query.map(|x| (x, x))
-                 .arrange_by_key_u()
+                 .arrange_by_key()
                  .join_core(&edges, |_n, &q, &d| Some((d, q)))
-                 .arrange_by_key_u()
+                 .arrange_by_key()
                  .join_core(&edges, |_n, &q, &d| Some((d, q)))
-                 .arrange_by_key_u()
+                 .arrange_by_key()
                  .join_core(&edges, |_n, &q, &d| Some((d, q)))
                  .filter(move |_| inspect)
                  .map(|x| x.1)
--- a/examples/bfs.rs
+++ b/examples/bfs.rs
@@ -113,8 +113,8 @@ where G::Timestamp: Lattice+Ord {
         let edges = edges.enter(&inner.scope());
         let nodes = nodes.enter(&inner.scope());
 
-        inner.join_map_u(&edges, |_k,l,d| (*d, l+1))
+        inner.join_map(&edges, |_k,l,d| (*d, l+1))
              .concat(&nodes)
-             .group_u(|_, s, t| t.push((*s[0].0, 1)))
+             .group(|_, s, t| t.push((*s[0].0, 1)))
      })
 }
--- a/examples/bfs.rs
+++ b/examples/bfs.rs
@@ -59,24 +59,18 @@ fn main() {
         println!("performing BFS on {} nodes, {} edges:", nodes, edges);
 
         if worker.index() == 0 {
-
-            // trickle edges in to dataflow
-            for _ in 0..(edges/1000) {
-                for _ in 0..1000 {
-                    graph.insert((rng1.gen_range(0, nodes), rng1.gen_range(0, nodes)));
-                }
-                worker.step();
-            }
-            for _ in 0.. (edges % 1000) {
+            for _ in 0 .. edges {
                 graph.insert((rng1.gen_range(0, nodes), rng1.gen_range(0, nodes)));
             }
         }
 
-        println!("loaded; elapsed: {:?}", timer.elapsed());
+        println!("{:?}\tloaded", timer.elapsed());
 
         graph.advance_to(1);
         graph.flush();
         worker.step_while(|| probe.less_than(graph.time()));
+
+        println!("{:?}\tstable", timer.elapsed());
 
         for round in 0 .. rounds {
             for element in 0 .. batch {
@@ -88,12 +82,12 @@ fn main() {
             }
             graph.flush();
 
-            let timer = ::std::time::Instant::now();
+            let timer2 = ::std::time::Instant::now();
             worker.step_while(|| probe.less_than(&graph.time()));
 
             if worker.index() == 0 {
-                let elapsed = timer.elapsed();
-                println!("{:?}:\t{}", round, elapsed.as_secs() * 1000000000 + (elapsed.subsec_nanos() as u64));
+                let elapsed = timer2.elapsed();
+                println!("{:?}\t{:?}:\t{}", timer.elapsed(), round, elapsed.as_secs() * 1000000000 + (elapsed.subsec_nanos() as u64));
             }
         }
         println!("finished; elapsed: {:?}", timer.elapsed());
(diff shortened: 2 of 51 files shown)


frankmcsherry:
Some not especially scientific measurements.

## BFS 

The `examples/bfs.rs` program does 1M updates to a G(n,m) random graph, with (n,m) either (1000, 2000) or (1M, 10M). The elapsed durations for various batchings (batch size x round) are:

| (1000, 2000) | `MergeBatcher` | `RadixBatcher` |
|----------------:|-------------:|---------:|
| 1000 x 1000 | 31.018s |  32.680s |
| 10000 x 100 | 33.781s | 35.445s |
| 100000 x 10 | 41.783s | 44.386s |
| 1000000 x 1 | 55.968s | 62.912s |

Most of the time in `(1000, 2000)` is spent in `group`, doing complicated logic that has nothing to do with arranging data. The differences here suggest that merge is a bit of an improvement over the course of the millions of updates (the bfs churns for a bit, so the 1M updates multiply).

| (1M, 10M) | `MergeBatcher` | `RadixBatcher` |
|----------------:|-------------:|---------:|
| 1000 x 1000 | 56.526s | 62.685 |
| 10000 x 100 | 42.028s | 44.797s |
| 100000 x 10 | 30.152s | 31.666s |
| 1000000 x 1 | 20.032s | 21.011s |

The `(1M, 10M)` configuration has relatively less computation (but more time in "random access"), as the updates don't propagate much and are relatively sparse. Much more of this time is spent arranging the 10M+ initial edges and 1M+ updates that come through.

## Difference

The `examples/difference.rs` program feeds updates as fast as it can at `consolidate`, which accumulates the entries in place. Both `RadixBatcher` and `MergeBatcher` do in-place updating, but it is more painful for `RadixBatcher` which just repeatedly sorts and compacts its buffers.

This example is a bit unfair in favor of the merge batcher, in that the data come in largely sorted, and the initial `sort_unstable` on each buffer will be relatively faster than on random data. `RadixBatcher` seems less able to take advantage of this post-radixsort.

The main parameter in `difference.rs` is the number of keys, which determines how much compaction can happen. The merge batcher should be able to take advantage of a small key space, whereas the radix batcher waits until it has doubled its allocation past 1 million (a hard-coded limit we need for practical reasons; if we wanted to improve perf we could lower the threshold to the number of keys).

The following measurements are in "elements per second", the sustained rate of insertion at the point the approach has inserted 100m elements.

| keys | `MergeBatcher` | `RadixBatcher` |
|-----------:|---:|---:|
| 1 | 20,342,042.97 | 10,217,973.38 |
| 1K | 12,545,745.72 | 9,775,039.40 | 
| 1M | 13,198,872.70 |  7,054,789.25 | 

The merge batcher wins here, but possibly unfairly due to the sortedness of the input data. Also, for small key counts (e.g. 1) the merge approach of compacting buffers first and then merging is much better than the radix approach of sorting first then compacting.

## Degrees

The `examples/degrees.rs` program computes the degree distribution for a random graph, essentially doing a count first over a sparse set, and then over a dense set. No compaction is possible, and while the set of keys is relatively sparse in the second aggregation, the distinct values being merged has no repetetion; if anything, this should ask the merge approach to consider more of the key than before.

I ran the open-loop measurements, capturing the median, 99th%, and maximum latencies over 10s. I did this for both 10k nodes and 50k edges, and for 10M nodes and 50M edges.

| (10k, 50k) | `MergeBatcher` | `RadixBatcher` |
|--:|--:|--:|
| median | 58,054ns | 79,853ns |
| 99th% | 728,298ns | 833,590ns |
| maximum | 3,323,825ns | 2,637,731ns | 
 
| (10M, 50M) | `MergeBatcher` | `RadixBatcher` |
|--:|--:|--:|
| median | 263,782ns | 2,173,084ns |
| 99th% | 10,736,017ns | 42,820,971ns |
| maximum | 60,146,207ns | 101,595,952ns | 

I think this looks generally better for merging, especially on the larger graph. I'm not sure what to make of the maximum measurement on the smaller graph; re-execution suggests that the median and 99th% are stable, but there is substantial variation for the maximum measurement (second run, 2.44ms and 2.73ms for merge and radix respectively).

It is still plausible that the radix sort saturates at a higher rate (100k elts / sec is not especially high); I don't have a great methodology to test that accurately at the moment.

frankmcsherry:
More measurements, this time using `tpchlike`. I've run through all the queries, using both merge and radix batching, with two different settings of logical and physical batching: `(1,1000)`  and `(1000, 1)`, corresponding to "distinct times, 1000 at a time" and "times / 1000".



| query | M(1,1000) | M(1000,1) | R(1,1000) | R(1000,1) |
|------:|----------:|----------:|----------:|----------:|
|   q01 |    1.594s |    0.962s |    1.961s |    1.171s |
|   q02 |    0.525s |    0.493s |    0.525s |    0.500s |
|   q03 |    1.589s |    1.335s |    1.673s |    1.476s |
|   q04 |    2.214s |    1.500s |    2.011s |    1.487s |
|   q05 |    2.682s |    3.100s |    2.764s |    3.117s |
|   q06 |    0.289s |    0.303s |    0.309s |    0.315s |
|   q07 |    2.793s |    2.830s |    3.330s |    3.049s |
|   q08 |    4.401s |    4.513s |    4.596s |    4.804s |
|   q09 |    7.523s |    7.042s |   10.940s |   10.894s |
|   q10 |    2.453s |    2.141s |    4.904s |    4.559s |
|   q11 |   11.405s |    3.543s |   12.134s |    3.989s |
|   q12 |    0.918s |    0.955s |    1.058s |    1.046s |
|   q13 |    2.091s |    1.938s |    2.269s |    2.049s |
|   q14 |    0.722s |    0.726s |    0.703s |    0.719s |
|   q15 |    1.653s |    1.725s |    1.829s |    1.836s |
|   q16 |    0.589s |    0.475s |    0.825s |    0.682s |
|   q17 |    3.742s |    3.783s |    3.754s |    4.003s |
|   q18 |   10.990s |    4.715s |   11.910s |    5.054s |
|   q19 |    0.467s |    0.466s |    0.510s |    0.510s |
|   q20 |    1.361s |    1.351s |    1.465s |    1.444s |
|   q21 |    8.080s |    6.397s |    8.749s |    7.084s |
|   q22 |    7.278s |    3.500s |    8.725s |    3.960s |

I think it is generally the case that the `M` measurements are better than the corresponding `R` measurements, with a few stand out examples being `q09`, `q10`, and `q16`. These queries all happen to use `count_total` in the radix formulation, because we are accumulating non-integer keys. There is a lot of room to gain there, because we avoid all the hashing and such. And, apparently, we don't lose too much on the queries where we were able to use `count_total_u`.

PR #101, opened 2018-04-16
remove unused getopts

qmx:
continuation of https://github.com/frankmcsherry/timely-dataflow/pull/150


frankmcsherry:
Thanks! No clue why that was still around. :)

PR #110, opened 2018-07-23
Fixes

siddharthasahu:


examples/arrange.rs
```diff
@@ -29,11 +29,11 @@ fn main() {
     let edges: usize = std::env::args().nth(2).unwrap().parse().unwrap();
     let batch: usize = std::env::args().nth(3).unwrap().parse().unwrap();
     let pre: usize = std::env::args().nth(4).unwrap().parse().unwrap();
-    let inspect: bool = std::env::args().position(|x| x == "inspect").is_some();
+    let inspect: bool = std::env::args().nth(5).unwrap() == "inspect";
```
frankmcsherry: This is different now. Not that it was pretty before, but now it will crash if you don't supple "not_inspect" or some other 5th argument. What I've been using elsewhere is `.any(|x| x == "inspect")`.

frankmcsherry:
Most of these look great, with the exceptions of the cases that now crash without a "inspect/else" argument supplied. If you want to swap those to
```rust
std::env::args().any(|x| x == "inspect")
```
it looks good to go.

examples/arrange.rs
siddharthasahu: I did want to highlight that the argument is fixed to a position and not `any`. How about something like: `if let Some(input) = std::env::args().nth(5) { input == "inspect" } else { false };`

frankmcsherry: Pondering this a bit more, let's break things and make it be a positional argument like you've suggested. The right answer is probably a heavier-weight arg-parsing crate, but for the simple examples probably best to be clear about how many arguments there are to know how many to strip for timely.

siddharthasahu: Makes sense. One can always use `no_inspect` to turn it off, for instance. 

However, the skip() part does not seem mandatory for timely? I am able to send (at least the -w) arguments to timely even without a skip() and timely does the right thing.

PR #120, opened 2018-11-08
Unstall arrangements

frankmcsherry:
This addresses (horribly) a bug reported by @comnik in which imported arrangements could report incomplete frontiers once imported, until the first non-empty batch is transmitted.

The fix here is *not great*, and really much of this should be overhauled once we have a better handle on how to unify the two ways in which batches flow around. 

Test `test_import_stalled_dataflow` added which fails on master but passes here.


comnik:
Happy to report that this works!

PR #126, opened 2018-12-28
Add import_named

### LINKED ISSUE #125: Use arrangement name on import
Currently, each trace import results in a source operator with the generic name "ArrangedSource" (https://github.com/frankmcsherry/differential-dataflow/blob/master/src/operators/arrange.rs#L275).

Would it be possible to stash an arrangement's name within the `TraceAgent` and make use of that in the source? E.g. replace "ArrangedSource" with "Source( ARRANGEMENT_NAME )"?

Another minor thing is that `arrange_by_self` and `arrange_by_key` can't pass on names to `arrange` yet, but I'd be happy with using `.map(...).arrange()` directly.

If the above makes sense I'm happy to submit a PR.

comnik:
Closes #125

### DIFF
--- a/src/operators/arrange.rs
+++ b/src/operators/arrange.rs
@@ -268,11 +268,20 @@ where T: Lattice+Ord+Clone+'static, Tr: TraceReader<K,V,T,R> {
     ///     }).unwrap();
     /// }
     /// ```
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Looks good! I think the travis failure is due to me having checked in a faulty Cargo.toml (odd, but not uncommon).

PR #134, opened 2019-02-05
Monoid and Abelian

frankmcsherry:
This shifts the `Diff` trait to be an alias of `Abelian`, corresponding to an Abelian (commutative) group. There is also a `Monoid` trait corresponding to a commutative monoid. `Abelian` is `Monoid` with the addition of the `Sub` and `Neg` traits.

In many cases `Monoid` is sufficient, and `Abelian` is only required where subtraction is needed. Operators like `join` do not require subtraction, though `antijoin` does in its current implementation.

More generally, we should think about shifting some of the uses of negation to other forms. Iteration requires negation to correct for its initial input, but this is only required when we have an initial input and not otherwise.

cc @eoxxs

### DIFF
--- a/examples/monoid-bfs.rs
+++ b/examples/monoid-bfs.rs
@@ -1,0 +1,173 @@
+extern crate rand;
+extern crate timely;
+extern crate differential_dataflow;
+
+#[macro_use]
+extern crate abomonation_derive;
+extern crate abomonation;
+#[macro_use]
+extern crate serde_derive;
+extern crate serde;
+
+
+use rand::{Rng, SeedableRng, StdRng};
+
+use timely::dataflow::*;
+use timely::dataflow::operators::probe::Handle;
+
+use differential_dataflow::input::Input;
+use differential_dataflow::Collection;
+use differential_dataflow::operators::*;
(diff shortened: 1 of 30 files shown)


frankmcsherry:
The commit just above introduces the `group_solve` method to the `GroupArrange` trait. This method takes a different `logic` closure, one which is presented with the current `input`, what would be the `output` if left unadjusted, and a buffer `updates` into which the logic should write any necessary diffs. For the standard `group_arrange` method this applies its logic to the input, puts the results in the update buffer, and then subtracts the current output, producing the necessary updates to correct to the intended output.

However, this method is more general, in that we can operate on collections whose difference type is only a commutative monoid, and not a commutative group (it may not have subtraction). For example, in something like shortest path computation this difference type may track the shortest path (min, rather than sum) and the update rule just considers whether we need an updated min as output, rather than attempting to subtract the old min from the new min (won't work).

A new `example/monoid-bfs.rs` demonstrates this in practice.

frankmcsherry:
That should be `MonoidVariable`, not `MonotonicVariable`, though .. similar in spirit.

PR #141, opened 2019-02-22
Monoid simplifications

frankmcsherry:
This PR removes various constraints from `Monoid`, and consequently from `Abelian` as well.

The new constraints are

```rust
pub trait Monoid : AddAssign + ::std::marker::Sized + Data + Clone
```

which replaces `Add` with `AddAssign` (based on actual use in the code, ability to re-use resources) and replaces `Copy` with `Clone`. The current implementation is fairly clone-eager, and should probably be revised in various places, and perhaps also require `AddAssign<&Self>`.

The `Abelian` requirements were relaxed to not include `Sub`, instead just using `Neg` everywhere.

@eoxxs

### DIFF
--- a/examples/monoid-bfs.rs
+++ b/examples/monoid-bfs.rs
@@ -28,13 +28,12 @@ pub struct MinSum {
     value: u32,
 }
 
-use std::ops::{Add, Mul};
+use std::ops::{AddAssign, Mul};
 use differential_dataflow::difference::Monoid;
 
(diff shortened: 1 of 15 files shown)


frankmcsherry:
Ah, the PR also adds a `DiffVector<R>` type which can contain a variable length sequence of `R` elements, where addition, negation, etc are done pointwise. The intent here is to support projects that would like `DiffPair` like functionality but need to pick a static update type.

PR #145, opened 2019-02-23
Remove `Lattice::maximum()`

frankmcsherry:
This PR removes the `maximum()` method from the `Lattice` trait, enabling a larger class of timestamps appropriate for differential dataflow. Previously, `maximum()` was used in only two places:

1. In `reduce.rs` where it simplified the maintenance of the running `meet`. Where we previously started from `maximum()` and went down, we now use `Option<T>` like we probably should have done from the start.

2. In `Lattice::advance_by()` we used `maximum()` as the return value for an empty frontier. The result should be an `Option<T>` but we are a bit too lazy to do that right now (the result is only optional with an empty input frontier, which can be observed by the caller and probably shouldn't result in a call.

The advantage of this relaxation is that we can use timestamp types without a largest element. The specific example that precipitated this was `Vec<T> where T: Lattice`.

### DIFF
--- a/experiments/src/bin/multitemporal.rs
+++ b/experiments/src/bin/multitemporal.rs
@@ -9,7 +9,6 @@ extern crate differential_dataflow;
 use rand::{Rng, SeedableRng, StdRng};
 
 use timely::dataflow::ProbeHandle;
-// use timely::progress::timestamp::RootTimestamp;
 
 use timely::dataflow::operators::unordered_input::UnorderedInput;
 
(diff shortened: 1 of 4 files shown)


frankmcsherry:
The most recent commit updates `Lattice::advance_by(frontier)` to update `self`, rather than force it to create a new timestamp element. The old behavior should be recoverable by cloning a time and advancing it, but this new version has the potential to avoid a substantial amount of timestamp allocation.

cc @comnik

comnik:
Great, thanks. Will report whether that plays nicely with the `enter_at` workaround.

PR #146, opened 2019-02-24
Unmanaged arrangements

comnik:
This isn't ready for prime-time yet, but I could use some feedback. 
If the general approach seems sane, I can clean things up.

For one, I haven't actually integrated this with declarative, to see whether it actually solves the unregistering problem and leads to a proper cleanup. Will check that now.

The other thing is that now `import_named` and `import_unmanaged` do mostly the same things. I think the former could be reduced to the latter + `Box::leak(killswitch)` but I'm not sure.
Other than that, current arrangement API is unaffected by these additions.

### DIFF
--- a/src/operators/arrange.rs
+++ b/src/operators/arrange.rs
@@ -327,6 +327,73 @@ where T: Lattice+Ord+Clone+'static, Tr: TraceReader<K,V,T,R> {
 
         Arranged { stream, trace }
     }
+
+    /// Imports a trace into an unmanaged arrangement, which can be
+    /// cleaned up by its user.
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Tbh, this looks pretty sane.

There is a bunch of horrible debt around trace importing and such, and while I should probably tidy that at up some point this doesn't seem to make things any worse.

Why `Box::leak` instead of `mem::forget`?

frankmcsherry:
Another option: both `Rc<RefCell<Option<..>>>` things could be strong references, and the killswitch could have a method `push()` that sets the option to `None`. This would allow uninterested people to drop the kill switch and not terminate the input, and only require complexity of the people who want to retain the ability to terminate the input.

PR #149, opened 2019-02-26
Import unset

frankmcsherry:
This PR steals from #146 (with permission) to introduce a method `import_core` that returns in addition to an arrangement a `ShutdownButton` (name pending) with a single method `push()` that causes the imported arrangement to cease producing output batches. This is meant to allow clean shutdown of dataflow graphs without requiring the termination of the input arrangements on which they depend.

The PR also uses `CapabilitySet<T>` from timely dataflow, which tidies up the implementation quite a bit!

cc @comnik @utaal

### DIFF
--- a/src/operators/arrange.rs
+++ b/src/operators/arrange.rs
@@ -313,9 +313,8 @@ where
                 if let Some(ref mut capabilities) = *capabilities {
 
                     let mut borrow = queue.1.borrow_mut();
-                    while let Some((frontier, sent)) = borrow.pop_front() {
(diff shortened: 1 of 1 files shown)


comnik:
This looks much tidier indeed! Didn't know about `CapabilitySet`. 

Re naming, I like `ImplementOfOurDemise` but could also live with `ShutdownHandle` / `ShutdownButton`. What I would prefer is something like `shutdown()` rather than `push()` because the latter always evokes communication in the dataflow context.

frankmcsherry:
Early reports are that this looks good, so we'll go with it!

PR #150, opened 2019-02-27
Fix a few small typos in the mdbook

benesch:


frankmcsherry:
These look good, thanks very much!

PR #152, opened 2019-03-05
Associated types for TraceReader

frankmcsherry:
This PR introduces associated types `Key`, `Val`, `Time`, and `R` for `TraceReader`. These types were previously generic parameters, and infected all of the types related to trace management. Many types are now substantially simpler to express, because the `TraceReader` implementation drives these four "associated" types.

As an example, the `arrange_named` method, the core of arrangement, now has the following signature:

```rust
    fn arrange_named<Tr>(&self, name: &str) -> Arranged<G, TraceAgent<Tr>>;
```

whereas both `Arranged` and `TraceAgent` used to have various `K, V, T, R` generic type parameters.

The only downside so far is that due to the lack of "implied bounds", there are some places where we must repeat various type bounds that are "known" to be true. For example, we may need to write

```rust
where
    Tr: TraceReader<Time=G::Timestamp>+Clone+'static,
    Tr::Batch: BatchReader<Tr::Key,Tr::Val,G::Timestamp,Tr::R>+'static,
    Tr::Cursor: Cursor<Tr::Key,Tr::Val,G::Timestamp,Tr::R>+'static,
```

where the second two constraints are constraints of the `TraceReader` trait itself, and so *should* already hold by definition. So, eyes open for that.

cc @comnik @ryzhyk

### DIFF
--- a/dogsdogsdogs/src/altneu.rs
+++ b/dogsdogsdogs/src/altneu.rs
@@ -75,7 +75,6 @@ impl<T: Timestamp> Refines<T> for AltNeu<T> {
 use differential_dataflow::lattice::Lattice;
 impl<T: Lattice> Lattice for AltNeu<T> {
     fn minimum() -> Self { AltNeu::alt(T::minimum()) }
-    fn maximum() -> Self { AltNeu::neu(T::maximum()) }
     fn join(&self, other: &Self) -> Self {
         let time = self.time.join(&other.time);
         let mut neu = false;
--- a/dogsdogsdogs/src/lib.rs
+++ b/dogsdogsdogs/src/lib.rs
@@ -11,6 +11,7 @@ extern crate serde;
 use std::rc::Rc;
 use std::collections::HashMap;
 use std::hash::Hash;
+use std::ops::Mul;
 
 use timely::PartialOrder;
 use timely::dataflow::Scope;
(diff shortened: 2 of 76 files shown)


comnik:
This looks great! I have to hold out for a bit before trying it out tho, I just got a bunch of stuff stabilised...

frankmcsherry:
No worries. It is pretty massively breaking, and wanted to get a read from you and @ryzhyk before doing anything, as it is probably pretty disruptive.

ryzhyk:
Looks neat, those `TraceReader` types were getting out of control :) Will definitely try it out, but probably in a day or two (it is going to affect a lot of my code, so I need to allocate a good chunk of time to it).

frankmcsherry:
One answer is certainly "this looks promising, but exhausting; please delay until next major release" which is certainly something I can do.

ryzhyk:
Depends on how much rush you're in to release 0.9. If you think it can wait a couple of days, I'm more than happy to do it.

ryzhyk:
@frankmcsherry, sorry, I'm still dealing with deadlines. Probably won't do it till next week.

frankmcsherry:
No worries! I've got a lot to do too. No rush to land this.

comnik:
So my stuff builds, tests pass, no problems there. Unfortunately my major touchpoint is the PrefixExtender-compatible index type, which maintains three different trace types (count, propose, validate).

This leads to some interesting ASCII art: https://github.com/comnik/declarative-dataflow/commit/c7d8e8d2aafb1ac44068882364a4b853203fd830#diff-b4aea3e418ccdb71239b96952d9cddb6R311

ryzhyk:
My tests compile and run. Type signatures are much simpler now, but the `where` clauses in my function declarations are kind of gigantic now, as pointed out in the PR.

PR #156, opened 2019-03-10
Utilities for working with DiffVector

eoxxs:
I came across this while implementing multiple aggregations over at 3DF.


frankmcsherry:
Looks good!

PR #157, opened 2019-03-11
Replace #[inline(always)] with #[inline].

rw:
As a maintainer of another high-performance Rust project (Rust FlatBuffers), I've learned that forcing the compiler to inline can be problematic. The Rust toolchain (LLVM in particular) can make better decisions than we can (usually) when deciding when and what to inline. This PR switches all #[inline(always)] attributes to #[inline], in accordance with Rust best practices.

See also https://github.com/TimelyDataflow/timely-dataflow/pull/247

### DIFF
--- a/src/difference.rs
+++ b/src/difference.rs
@@ -24,7 +24,7 @@ pub trait Monoid : for<'a> AddAssign<&'a Self> + ::std::marker::Sized + Data + C
 	/// This is primarily used by differential dataflow to know when it is safe to delete an update.
 	/// When a difference accumulates to zero, the difference has no effect on any accumulation and can
 	/// be removed.
-	#[inline(always)]
+	#[inline]
 	fn is_zero(&self) -> bool { self.eq(&Self::zero()) }
 	/// The additive identity.
 	///
@@ -43,15 +43,15 @@ pub trait Abelian : Monoid + Neg<Output=Self> { }
 impl<T: Monoid + Neg<Output=Self>> Abelian for T { }
 
 impl Monoid for isize {
-	#[inline(always)] fn zero() -> Self { 0 }
+	#[inline] fn zero() -> Self { 0 }
 }
 
 impl Monoid for i64 {
-	#[inline(always)] fn zero() -> Self { 0 }
+	#[inline] fn zero() -> Self { 0 }
 }
 
 impl Monoid for i32 {
-	#[inline(always)] fn zero() -> Self { 0 }
+	#[inline] fn zero() -> Self { 0 }
 }
 
 /// The difference defined by a pair of difference elements.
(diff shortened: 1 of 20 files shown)


rw:
(Of course, if you find this isn't a good idea, feel free to disregard!)

frankmcsherry:
I'm testing it at the moment. No obvious regressions (or improvements) what I'm looking at, but what I'm looking at is a lot of LLVM not inlining things (both with and without the PR). 

| time | percent | self | symbol |
|:-----|:---|:----|:----|
| 1.09 s |   4.7% |	1.09 s	 	| core::cmp::impls::_$LT$impl$u20$core..cmp..Ord$u20$for$u20$u32$GT$::cmp::h575d60437fbfbd5d |
| 2.14 s  |  9.3%	 | 712.00 ms	 |	 _$LT$differential_dataflow..operators..join..Deferred$LT$K$C$$u20$V1$C$$u20$V2$C$$u20$T$C$$u20$R1$C$$u20$R2$C$$u20$R3$C$$u20$C1$C$$u20$C2$C$$u20$M$C$$u20$D$GT$$GT$::work::hbca5b27e480947e9 |
| 2.45 s  | 10.6%	| 703.00 ms	 |	 _$LT$differential_dataflow..trace..implementations..merge_batcher..MergeSorter$LT$D$C$$u20$T$C$$u20$R$GT$$GT$::merge_by::h5e6928cb34e35302 |
| 3.11 s  | 13.5% |	653.00 ms	 	| _$LT$differential_dataflow..trace..implementations..merge_batcher..MergeSorter$LT$D$C$$u20$T$C$$u20$R$GT$$GT$::merge_by::hff6e8aa31868167e |
| 592.00 ms  |  2.5% |	592.00 ms |	 	 core::ptr::write::h320521e7fe11e074 |
| 549.00 ms  |  2.3% |	549.00 ms |	 	 core::ptr::write::h424ab19d22d71db1 |
| 533.00 ms  |  2.3% |	533.00 ms |	 	 _platform_memmove$VARIANT$Haswell |
| 494.00 ms  |  2.1% |	494.00 ms |	 	 core::ptr::_$LT$impl$u20$$BP$mut$u20$T$GT$::offset::hc58867ee7a8dce5e |
| 1.74 s |    7.5% |	428.00 ms	 |	 _$LT$differential_dataflow..trace..implementations..ord..OrdValMerger$LT$K$C$$u20$V$C$$u20$T$C$$u20$R$GT$$u20$as$u20$differential_dataflow..trace..Merger$LT$K$C$$u20$V$C$$u20$T$C$$u20$R$C$$u20$differential_dataflow..trace..implementations..ord..OrdValBatch$LT$K$C$$u20$V$C$$u20$T$C$$u20$R$GT$$GT$$GT$::work::hccfbc0c5fc18e53e |

rw:
@frankmcsherry Maybe it would be clearer to start with TD?

frankmcsherry:
Maybe, but all of these "probably should have been inlined" methods are differential logic. I'm not sure that's the root of the problem. I probably need to learn more about how Instruments reports things (it is also surfacing methods that have `#[inline(always)]` attributes, so .. I don't know what's going on.

frankmcsherry:
Technically I am on vacation for two weeks, so maybe not doing anything for a while! :D

frankmcsherry:
The SOSP deadline just passed, which was blocking this (didn't want incidental performance regressions while under a deadline). Per the timely dataflow reasoning, this seems solid to merge.

PR #158, opened 2019-03-13
Mdbook updates

frankmcsherry:
Updates in response to confusions reported over in https://github.com/TimelyDataflow/timely-dataflow/issues/248. These include:

1. Making the examples run as actual binaries (as described by the text).
2. Add a size argument initially, to avoid the confusion of adding it later.
3. Elaborate on `concat` and `consolidate`.
4. Swap the order of the `query` pair for arrangements.


dhalperi:
LGTM. The actual output from `cargo` still pretends you're running `--example hello` [0], but I know how to gloss over that.

[0] E.g., https://github.com/TimelyDataflow/differential-dataflow/pull/158/files#diff-7e95bf10b9793c75249b784f913c404eR63

PR #160, opened 2019-03-19
hack to prevent panic

frankmcsherry:
This PR attempts to prevent a panic in which two already-advanced traces are joined. One will attempt to get a cursor for the other at frontier `Vec<G::Timestamp::minimum()>`, and will fail because the other trace has already been `distinguish_since`d something larger.

However, this really should have just been "nothing done, because such an early frontier means no data". So, the join operator tracks options of frontiers now, and doesn't request cursors that it can't have.

This is technically a hack because we could plausibly receive only a subset of batches in our first join execution, and this would cause a bunch of panics for similar reasons (requesting cursors up to boundaries that are no longer maintained). The longer-term fix should be to switch to a "bookmark" like implementation of `distinguish_since`.

@comnik is this anything you've seen? It was a quickly observed panic for me in `interactive`: just create two collections: put some data in and advance them a bit (to move the distinguish-since boundary past `vec![0]`) and then attempt to join them.

### DIFF
--- a/src/operators/join.rs
+++ b/src/operators/join.rs
@@ -313,8 +313,8 @@ impl<G, K, V, R1, T1> JoinCore<G, K, V, R1> for Arranged<G,K,V,R1,T1>
         let mut trace2 = Some(other.trace.clone());
 
         // acknowledged frontier for each input.
-        let mut acknowledged1: Option<Vec<G::Timestamp>> = None;//vec![G::Timestamp::minimum()];
-        let mut acknowledged2: Option<Vec<G::Timestamp>> = None;//vec![G::Timestamp::minimum()];
+        let mut acknowledged1: Option<Vec<G::Timestamp>> = None;
+        let mut acknowledged2: Option<Vec<G::Timestamp>> = None;
 
         // deferred work of batches from each input.
         let mut todo1 = Vec::new();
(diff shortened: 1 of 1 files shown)


comnik:
Never seen that, IIRC. I might be confusing the two frontiers again, but I always `distinguish_since(&[])` upon creation of a trace. Or are you talking about something different?

frankmcsherry:
Right, so imagine you `distinguish_since(&[])` two traces, feed them some data, advance them, feed data, make a bunch of batches, and then .. `join_core` them.

I think in this case there should be a panic on master. One of them will attempt `cursor_through(&[0])` as join starts each input "acknowledged" up to `&[0]` or whatever the minimum value is. However, with a bit of merging, this frontier is no longer available for `cursor_through`.

comnik:
Uh I think I know why I haven't encountered that... It took me an embarrassingly long time to realise, but the way the synthesis works, forcing everything along the query plan tree into this `CollectionRelation` thingy, is breaking re-used collections down via `as_collection` so the actual arrangement isn't even used. Could that explain it?

(I've introduced an `ArrangedRelation` to fix this, but it needs some more thinking, because the arrangement is only re-usable if its on the right key)

frankmcsherry:
That would explain not seeing this, I think!

The `interactive` project is happily re-using arrangements, but was exploding right away. :D

comnik:
Although (in my defense) `Hector` _is_ re-using the arrangements and I haven't seen it there either. Might not be applicable there? I was going to ask you about the various trace frontier modifications within the extender implementation anyways, I don't really follow there...

frankmcsherry:
It's literally in the `join_core` logic, so Hector probably doesn't surface the same issue.

comnik:
In that case I better get to fixing the re-use problem :P

PR #165, opened 2019-04-02
group to reduce

li1:
Some minor adjustments for the mdbook after the name change from `group` to `reduce`


frankmcsherry:
Looks good, and thanks very much!

PR #167, opened 2019-04-10
Add arrange_core

comnik:
This is mostly intended as a place to keep track of the discussion around "trusted arrange" / arrange without exchange baked in.

The constraints on the `Arrange` trait got hardened to include `K: Data+Hashable` and `V: Data`, which was practically the case anyways (if I'm not missing something). But if arrangements are made to work without any shuffling, it might be nice to loosen those again?

The more interesting question is about `lookup`. I think it makes sense to have a separate `lookup_broadcast` and leave `lookup` the way it is (thus entrusting users to only use it on sufficiently exchanged arrangements).

### DIFF
--- a/src/operators/arrange.rs
+++ b/src/operators/arrange.rs
@@ -26,7 +26,7 @@ use timely::dataflow::operators::{Enter, Map};
 use timely::order::{PartialOrder, TotalOrder};
 use timely::dataflow::{Scope, Stream};
 use timely::dataflow::operators::generic::{Operator, source};
-use timely::dataflow::channels::pact::{Pipeline, Exchange};
+use timely::dataflow::channels::pact::{ParallelizationContract, Pipeline, Exchange};
 use timely::progress::Timestamp;
 use timely::progress::frontier::Antichain;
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Oh nice. I was thinking about doing this tomorrow. Could save me the trouble! :D

frankmcsherry:
Could we remove the `Hashable` trait requirement? Or rather, add it only to `arrange_named`?

comnik:
It should be possible but actually I don't know how to express that in Rust. I think I can't just give an additional `K: Hashable` constraint here: https://github.com/TimelyDataflow/differential-dataflow/blob/ed8aef05dd349913084a63abcfaaecb934c109e4/src/operators/arrange.rs#L850

I will try again tomorrow!

frankmcsherry:
Yup, you should just be able to add it to the `where` constraints in relevant methods!

comnik:
Yeah I should head to bed... It was complaining about `arrange`, which of course then also needs the constraint. Seems to work now.

PR #173, opened 2019-05-06
Unbreak dogsdogsdogs and pin Differential version

comnik:
This uses the new ExchangeData and pins the most recent Differential revision to prevent the examples from breaking in the future.

### DIFF
--- a/dogsdogsdogs/Cargo.toml
+++ b/dogsdogsdogs/Cargo.toml
@@ -11,6 +11,6 @@ abomonation = "0.7"
 abomonation_derive = "0.3"
 timely = { git = "https://github.com/TimelyDataflow/timely-dataflow" }
 timely_sort="0.1.6"
-differential-dataflow = { path = "../" }
+differential-dataflow = { git = "https://github.com/TimelyDataflow/differential-dataflow", rev = "443ec7c66752cd00cc90c2e3b8791a47af6b6d7f" }
 serde = "1"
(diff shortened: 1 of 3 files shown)


frankmcsherry:
Ah oops. I think I went and unbroke it (not the examples) without this PR, for which I apologize. Does the current version (master) work for you, or were there other unbreaking changes you made? 

I'm not sure I grok the version pinning; can you explain what is up with that?

comnik:
No worries, should be fine! We were just using the examples as computations for testing the remote log analysis.

With the current revision (for which we know things build) pinned in the dogsdogsdogs Cargo.toml, the examples should continue working even as master introduces breaking changes. The alternative might be running dogsdogsdogs tests on Travis with each build? Just so that people interested in trying them out have a smooth experience.

frankmcsherry:
I see. Let me fix up the examples, and look in to extending the CI. @benesch rigged this for timely, and I bet I can copy/paste things for differential too. :D

comnik:
:D Feel free to close this whenever.

frankmcsherry:
I'll leave it open until I sort out the CI thing, which I haven't fixed yet. :)

frankmcsherry:
@comnik I suspect this is past relevance; does that sound right, or is there a lingering issue?

comnik:
@frankmcsherry no I think this is definitely past relevance :)

PR #174, opened 2019-05-06
Fix use of cursor through in join

comnik:
I tried querying a remote Timely computation with 3DF today. Sourcing the remote events was easy enough, but as soon as I start compacting traces I run into the cursor_through problem again. I'm not sure what is happening, but one input contains a batch with upper say at [100ms], whereas the trace is already compacted (advanced  + distinguished) up to something like [2s].

Looking at how the acknowledged frontiers were updated in join, I came up with this fix.

The idea is to initialise both acknowledged frontiers at the respective distinguish frontiers and then replace them as soon as any batch upper is in advance of that trace.

This problem is not in any way critical right now and the fix should probably be scrutinised a bunch...

[edit: dangit, this branch now includes the dogsdogsdogs PR as well, sorry. the relevant changes are [here](https://github.com/TimelyDataflow/differential-dataflow/pull/174/files#diff-7a261183716f8784618fda73a689b19d)]

[edit2: uh oh, tests are failing...]


comnik:
Second attempt. This has the distinct advantages of a) being a much smaller change, b) not breaking all the tests.

frankmcsherry:
This is definitely a compact change! I'll grab some time tomorrow and work through what is going on here; sorry for the delay!

comnik:
No rush, I just couldn't leave it broken like that...

My mental model is analogous to the forwarding by advance frontier on importing a trace. When looking at advanced traces, we must look at all times forwarded by the distinguish frontier. So an equivalent fix should be advancing `batch.upper()` appropriately.

I'll update this once I figure out how to reproduce in a unit test.

frankmcsherry:
Looking at this, it looks like it perhaps plasters over a larger problem. At least, if the `acknowledged` frontiers exist but do not line up, we risk double or under counting outputs.

Let me talk through what is supposed to happen, and we can perhaps try and sort out a more robust solution (or determine that this actually works great).

Informally, the `join` operator has some internal state that is the pair of frontiers
```rust
(acknowledged1, acknowledged2)
```
The intent of these frontiers is that the join has either produced, or has enqueue the production of all tuples from the prefix of batches up to and including the respective `acknowledged` frontiers.

Again informally, each batch that is received logically advances one of these two frontiers to the batch's upper frontier, and joins all contained tuples in the batch with the *other* trace up through its acknowledged frontier. Once a batch is fully processed, we can advance its acknowledged frontier to its upper frontier and inform its trace that we will only need cursor access from that point on.

By all accounts (correct me if wrong) this seems to work great once things are up and running. However, we need to pay some attention to the case where we start with some traces that have perhaps already begun to merge batches.

The current logic allows each `acknowledged` to be an `Option<Frontier>` which is initially `None` indicating that no batches have been read yet. This is meant to represent the frontier `[0]`, which contains no data, but we possibly aren't allow to ask about that frontier (if we have advanced the capability). Instead, as long as an `acknowledged` is `None` we short-circuit `cursor_through` and just do no work instead. As soon as we start seeing batches, we start setting `acknowledged` to `Some(batch.upper())`.

As a `join` starts up, essentially for the first time, it drains each of its inputs using the logic above. It pulls batches out, enqueues work, and first sets and then updates `acknowledged`, for both of its inputs. 

The hope (perhaps flawed) is that in the first invocation this drains `input1` and sets `acknowledged1`, but enqueues no work. It then drains `input2`, sets and updates `acknowledged2`, and enqueues work with `cursor_through(acknowledged1)`. The very strong hope is that this is an acceptable call to make, in that we have not merged past `acknowledged1` yet, which is motivated by the belief that we have just been shown all batches that exist, and the upper frontier of the very last of these is a legit frontier to find.

frankmcsherry:
This is all very tenuous reasoning, and I'm up for the idea that it should just be fixed (e.g. with `Bookmark` tokens that go out with each batch and get used in `cursor_through`).

I would like to understand the case you have, and where the reasoning is flawed, but perhaps the right thing to do it tear it all down and do it better.

comnik:
> The very strong hope is that this is an acceptable call to make, in that we have not merged past acknowledged1 yet, which is motivated by the belief that we have just been shown all batches that exist, and the upper frontier of the very last of these is a legit frontier to find.

I can follow until here. I don't quite see why either a) the last batch's upper would necessarily line up with the distinguish frontier of its trace or why b) the last batch's upper on trace A would be safe to use in acquiring a cursor on trace B?

Is the assumption that distinguish frontiers for *all* traces anywhere in the system advance in lockstep? And wouldn't I (meaning user-level code) then have to replace `distinguish_since(frontier)` with `distinguish_since(intersection(frontier, last_batch.upper()))`?

I would however also very much like to understand why it breaks in my case. It's quite hard to tease out, because it only occurred when replaying a remote Timely log stream into 3DF and installing a query a little later. I suspect that those two computations not living on the same real-time might have to do something with the problem. The only thing I do know is that I ended up with two traces, one forwarding a batch whose `upper` was way behind the other's distinguish frontier.

frankmcsherry:
I'm pretty sure the code has drifted a bit underneath this PR; I think things are probably fixed (at least, not hearing about any related crashes). Feel free to re-open if this is still a fix to an extant problem.

PR #176, opened 2019-05-11
Trace batch tidy

frankmcsherry:
This PR is trying to add a bit more discipline to how traces are built from batches. Previously, a trace would contain all non-empty batches, but empty batches didn't quite make it to the trace until there were some records to transmit and .. it was all a bit hack-y with some unfortunate consequences.

The PR also re-organizes `operators/arrange.rs` into several files, in part so that each of the moving parts more obviously should have independently specifiable behaviors.

The `TraceWriter` accepts a sequence of batches, whose `lower` and `upper` frontiers should line up. It also has a `seal` method which helps submit an empty batch up to a supplied frontier. These batches go to the shared trace (if it exists), and to each of the private queues associated with imports.

The `TraceAgent` implements the `TraceReader` trait and supports the importing of traces into other dataflows. This does a bit of a dance with the existing trace: reading its current batches out, pushing them into a new private queue, and then dropping that queue into the list of the `TraceWriter`. The `import_core` method is its core functionality.

The `Arrangement` is the arranged analogue of a `Collection`, and has lots of methods that allow you to use it in weird ways.

---

The PR should be mostly functionality *preserving*, with a few new features: a trace now has a `read_upper` method that allows you to read out its current upper frontier (previously, there was some dead time until non-empty batches arrived).

There is the risk that the PR has some performance regressions, as we are now moving around more empty batches. We will have to check for that. Nothing requires the internals of the trace to actually move these many batches around, but we do now need to respect the contract that having shown the trace some empty batches it properly reflects that information.

### DIFF
--- a/src/operators/arrange/agent.rs
+++ b/src/operators/arrange/agent.rs
@@ -120,7 +120,6 @@ where
             .map_batches(|batch| {
                 new_queue.push_back(TraceReplayInstruction::Batch(batch.clone(), Some(Default::default())));
                 upper = Some(batch.upper().to_vec());
-                // new_queue.push_back((vec![Default::default()], batch.clone(), Some(Default::default())));
             });
 
         if let Some(upper) = upper {
--- a/src/operators/arrange/agent.rs
+++ b/src/operators/arrange/agent.rs
@@ -47,6 +47,7 @@ where
 
     type Batch = Tr::Batch;
     type Cursor = Tr::Cursor;
+
     fn advance_by(&mut self, frontier: &[Tr::Time]) {
         self.trace.borrow_mut().adjust_advance_frontier(&self.advance[..], frontier);
         self.advance.clear();
@@ -280,9 +281,6 @@ where
     {
         let trace = self.clone();
 
-        // Capabilities shared with a shutdown button.
-        // let shutdown_button = ShutdownButton::new(capabilities.clone());
-
         let mut shutdown_button = None;
 
         let stream = {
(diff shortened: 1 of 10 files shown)


frankmcsherry:
One additional perk here, which might help width debugging #174, is more clarity on what needs to be true for `cursor_through`, and that all batches that get handed out have made their way through the trace so that any `distinguish_since` failures should be more debuggable. Maybe.

frankmcsherry:
@comnik I'm planning on landing this soon, to expose a more current `upper` for each trace, but there might be some performance fall out. It should be correctable, if and once we see it, so do holler if you see any regressions. Unfortunately, the repo doesn't really have its own performance tests... =/

comnik:
Great! I'll have to read a bit more to understand what has changed and see whether there are performance regressions on my end.

frankmcsherry:
The main changes *should* mainly be that traces are force-fed more empty batches, where previously traces didn't really receive empty batches and everyone else had to work around that, looking at timely frontiers to pick up that information. Now both traces and `import` folks see empty batches, which allows traces to report an accurate "what is the most current `upper` I've seen" where it couldn't before.

The additional empty traces could gunk up the works for e.g. merging, but in principle we can drop them immediately once received and fix that problem, but we need to remove some consistency checks (e.g. that one batch.upper equals the next batch.lower, which needn't be true if there were empty batches in between that we dropped).

comnik:
Ah I understand now. Before reading the patch I though this would already do away with the distinguish frontier. I don't have a clear cut way to check for performance regressions either, but establishing one won't hurt, sooner or later.

Do you think `Arrange` could be publicly exported through `arrange/mod` as well? That would make this less of a breaking change.

frankmcsherry:
Ah crud! Fix incoming.

comnik:
`ShutdownButton` is not where it used to be either, although I'm not sure whether that is intended to be a public thing just yet?

frankmcsherry:
I just re-`pub use`d that back into `arrange` too. It was in `arrange::agent`. Gah, testing these things is ... :)

comnik:
No more breakage now :) My Travis setup should catch a bunch of them, as long as we're building off of Differential master.

PR #177, opened 2019-05-13
Enrich shutdown button

frankmcsherry:
This PR adds a `press_on_drop` method to `ShutdownButton`, which changes its behavior to shut down an import when the button is dropped. If this method is not called then dropping the button will not shut down the import.

@comnik

### DIFF
--- a/src/operators/arrange/agent.rs
+++ b/src/operators/arrange/agent.rs
@@ -329,6 +329,7 @@ where
 
 /// Wrapper than can drop shared references.
 pub struct ShutdownButton<T> {
+    on_drop: bool,
     reference: Rc<RefCell<Option<T>>>,
     activator: Activator,
 }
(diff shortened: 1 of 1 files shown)


comnik:
I have no strong opinions here. From 3DF experience, I'd say implementing "press-on-drop" within a different `ShutdownHandle` type might be an interesting alternative and could serve the additional purpose of bundling many shutdown buttons into a shutdown handle for an entire plan.

I'm a bit worried that the solution in this PR could lead to confusion when dealing with multiple shutdown handles, not knowing which of them will be pressed on drop and which won't.

utaal:
> I'm a bit worried that the solution in this PR could lead to confusion when dealing with multiple shutdown handles, not knowing which of them will be pressed on drop and which won't.

My two cents: I’m a fan of using a separate type for representing these kinds of different states. What about a ‘fn press_on_drop(self)’ that consumes self and returns a ShutdownButtonOnDrop (or similar) instead of mutating in place?

frankmcsherry:
Multiple types makes a lot of sense! Tbh, the PR was about floating an idea, and I'm not 100% clear on whether it is a good idea at all. Though, I now 100% agree that one should change the type if one wants a press_on_drop button! :D

frankmcsherry:
I'm going to go ahead with this to try it out. I think it simplifies some stuff we have, and with the type changed I'm less concerned about footguns elsewhere (it seems like you would have to opt in to the footgun, at least).

PR #179, opened 2019-05-27
[WIP] Simplify dogs^3

frankmcsherry:
This PR is the beginning of some simplifications / unifications of elements of dogsdogsdogs.

The first thing the PR does is remove the use of `Rc`, which previously existed because closures could not derive `Clone`. They seem to be able to do so now, so all is good. This has the potential to remove a bit of boxing, but also just tidies up some line noise.

The next thing the PR starts to do is to unify the `count`, `propose`, and `validate` logic in one operator that is effectively a streaming look-up by key and flatmap over values. This is probably very similar to the `lookup` method for `Arranged` in the core library, and perhaps these should be unified as well.

The main changes are that operators take a bit more programmatic structure (closures) telling them where to find keys in their inputs, how to behave on each matched input and values, and what sort of output to produce. The operator structures are otherwise nearly identical. 

One minor difference is that as both `count` and `validate` are 1:1, they perhaps shouldn't be forced to clone their input in the same way that `propose` must. Will investigate!


frankmcsherry:
@comnik

frankmcsherry:
I suppose another awkward moment is that `count` doesn't react to any associated value, but rather the accumulated `isize` count. It's not so hard to surface this too, but things are getting a bit weirder now...

frankmcsherry:
Though, surfacing the diff may be a fine way to allow a bit more sharing. Iirc, one of the awkward bits with this code is that we need distinct inputs in some places, and not in other places. If we allowed the closure to determine if diffs were multiplied, or just tested for non-zeroness, or whatever, then in principle one could use fewer traces (e.g. avoiding otherwise mandatory `distinct()` calls).

@comnik do you find that you are doing `distinct()` calls that you would prefer not to be doing?

frankmcsherry:
I'm thinking of an interface that looks a bit like this:

```rust
pub fn lookup_map<G, D, R, Tr, F, DOut, ROut, S>(
    prefixes: &Collection<G, D, R>,
    arrangement: Arranged<G, Tr>,
    key_selector: F,
    output_func: S,
) -> Collection<G, O, Tr::R>
where
    F: FnMut(&D, &mut Tr::Key),
    S: FnMut(&D, &R, &Tr::Val, &Tr::R) -> (DOut, ROut),
    ...
```

where a general collection has the ability to look into a general arrangement, as long as it is able to explain how to populate a key (`key_selector`) and what to do with any matching values and accumulated differences found.

utaal:
cc/ @li1 who's using the dogs^3 on master as a test case for Snailtrail

li1:
We just talked about this today, great stuff!

frankmcsherry:
Quick comments, interested in feedback from @comnik and @eoxxs 

There is now a method 

```rust
pub fn lookup_map<G, D, R, Tr, F, DOut, ROut, S>(
    prefixes: &Collection<G, D, R>,
    arrangement: Arranged<G, Tr>,
    key_selector: F,
    output_func: S,
) -> Collection<G, DOut, ROut>
where
    F: Fn(&D, &mut Tr::Key)+Clone+'static,
    S: Fn(&D, &R, &Tr::Val, &Tr::R)->(DOut, ROut)+'static,
```

which allows you to specify functions for key extraction and for output selection. This is "BYOMonoid", in that we ask you to perform the appropriate action between `R` and `Tr::R` to produce a `ROut`.

This means!!!

1. When in `count` we examine the `Tr::R` but retain `R` in the output,
2. When in `propose` we multiply the two to get `ROut`,
3. When in `validate` we ignore the `Tr::R`, but should perhaps take the minimum, if doing intersection?

Doing proper WCOJ, we would probably want `propose` to not multiply, if we haven't pre-performed a `distinct()`. In delta queries we do want the multiplication in `propose`. In `validate` I'm not sure what we are supposed to do, but probably not multiply (it is meant to be an intersection, not a join).

Anyhow, the question is mostly: given this degree of freedom, and the need to pay a bit more attention to what operation is done, do we have confidence we know what the right answers are? Should dogs3 be opinionated here, and what opinions should it have?

comnik:
This all looks very promising.

> closures could not derive Clone. They seem to be able to do so now, so all is good.

Hell yeah, this might also reduce the clutter for `AltNeu` wrapping...

> @comnik do you find that you are doing distinct() calls that you would prefer not to be doing?

I'm doing `distinct`'s on all cardinality-many attributes, which I would be very happy to get rid of. I used to distinct the concatenated output stream of all delta pipelines, dropped that recently, but not 100% sure that is safe to do.

> Doing proper WCOJ, we would probably want propose to not multiply,

Let me verify my understanding here: on non-distinct inputs, we might multiply prefix diffs by something > 1. That in itself would be correct, but it might happen at each extension stage, resulting in completely off output multiplicities. However not multiplying at all would also not be correct, right? So by "If we allowed the closure to determine if diffs were multiplied" you are referring to additional logic to decide at each stage whether multiplication is necessary?

> When in count we examine the Tr::R but retain R in the output,

If I understand correctly, this means we could for example use a pre-aggregated collection (e.g. with `SumProduct` diffs) and re-interpret the aggregated diffs as e.g. `isize`? @eoxxs This was one of the pain points in your FAQ work, IIRC?

I'll try and carve out some time to get `Hector` working with this branch. 

---

Semi-related: could distinct inputs be modelled as a separate diff type? E.g. an `Option<()>`, where `Some(()) + Some(()) = Some(())` for the monoid case, and some weird signed version of that for retractions. While I think those would work when looking at a fully accumulated state at some point in time, I find it a bit scary to think through what happens in general. For example [in lookup_map](https://github.com/TimelyDataflow/differential-dataflow/blob/bd2cae5e714f091081a1cbb51829a16b7afd8c74/dogsdogsdogs/src/operators/propose.rs#L145) shouldn't this then be equivalent to a `distinct()`-ed input?

eoxxs:
That seems pretty interesting!


"Doing proper WCOJ, we would probably want propose to not multiply, if we haven't pre-performed a distinct()."
I suppose indeed if we did not `distinct()` the collection before multiplying in `propose`, we would propose more than we said we counted, as the `count` trace is distinct. So `distinct` sounds good.
But I am not sure...

@comnik Yeah kind of :)

frankmcsherry:
> So by "If we allowed the closure to determine if diffs were multiplied" you are referring to additional logic to decide at each stage whether multiplication is necessary?

Yes, the meaning I guess being that in WCOJ one wants to avoid multiplication in `propose`, because further steps will add the additional attributes in (that accumulate up the count when they are projected out), and in vanilla delta join one wants the multiplication because this is the only time we'll see the relation and it is proposing all of its extensions and we want to capture the multiplicity.

> I'll try and carve out some time to get Hector working with this branch.

Any word here? No rush, but was looking at either merging or moving forward and up for any feedback! :D

> Semi-related: could distinct inputs be modelled as a separate diff type?

Yes, in principle, but in practice there is a minor annoyance. Each type is obliged to have a zero from which accumulations start, and against which "zero-ness" is tested to allow garbage collection. The `()` type has only one value. We could claim it is a zero, and then have `is_zero()` always return false, which is a mind-bender but could plausibly work? Probably the "right answer" is for most of the math to use `Option<Diff>` types for math, which .. should actually be zero overhead for `Diff = NonZeroFoo`. Future work.

frankmcsherry:
I've just added a `propose_distinct`, which does what it says. Each map preserves the count of the incoming prefix, and does not multiply it by the multiplicity of each extension.

frankmcsherry:
Well, let's see how it works. Report any issues! :D

PR #193, opened 2019-07-26
[WIP] Semigroup structure

frankmcsherry:
This PR is the beginning of a shift from `Monoid` to `Semigroup`.

The distinction between the two is that a semigroup need not have a zero element, which allows for a few new and important difference types. Perhaps the most important one, for relational processing, is a difference type wrapping `()` which simply means "present", without a multiplicity. Combined with a timestamp type of `()` this means that we needn't maintain difference structures for single timestamp, set-valued data, and need only maintain a timestamp for insert-only relational computations.

The main change in the PR lies in rewriting the `consolidate` methods to carefully avoid using the `zero()` method offered by the `Monoid` trait. These algorithms never *needed* to use the zero element, it was just a convenient way to start accumulations. It is now less convenient.

There are several places where these is potential performance fall-out, due to LLVM's inability to optimize out some "obvious" array bounds work (which are introduced because ... Rust). These are near the `split_at_mut` calls, which are the only safe way we have to convince Rust that we are working with non-overlapping slices. 

More performance testing is required before shipping this, but it passes all tests.


frankmcsherry:
As another example of the potential use, we presently have to write for the `MinSum` monoid (which tracks the minimum of values under + and takes their sum under x, to track minimum length paths):
```rust
impl Monoid for MinSum {
    fn zero() -> MinSum { MinSum { value: u32::max_value() } }
}
```
This isn't really correct, as this is not a zero for the monoid (it is close, and it would leave any other element undisturbed if used for addition). If these values got introduced, they would be immediately discarded as "equal to zero".

Instead, we write
```rust
impl Monoid for MinSum {
    fn is_zero(&self) -> bool { false }
}
```
which just indicates that any instance of `MinSum` is non-zero. Elsewhere in the code, we avoid creating such zero elements and reasoning using them.

comnik:
Somewhat off-topic: how do you come up with statements such as

> LLVM appears to struggle to optimize out Rust's split_at_mut, which would prove disjointness 
> using run-time tests.

Do you check the bytecode manually?

frankmcsherry:
Kinda. I went to https://godbolt.org which will show you the compiled code, but the Rust playground will also show you some stuff (a bit harder to read it there, imo). Benchmarking revealed minor differences, but not epic amounts. I should probably double check, as unsafe code is gross (even though this is relatively minor).

comnik:
Thanks, didn't know about godbolt!

frankmcsherry:
Okies; landing this to get it some exercise. No known defects, though a touch of `unsafe` code in `consolidation.rs` (to get around Rust and LLVM not cooperating on `split_at_mut()`).

PR #195, opened 2019-08-03
Example for item-based collaborative filtering

sscdotopen:
This is a somewhat realistic example of item-based collaborative filtering with Jaccard similarity.


frankmcsherry:
Interesting! Thanks very much!

I don't have a good read on the performance (whether this is meant to go fast, or to be clear, or a different goal) but there is at least one improvement that can cut out one arrangement: the 
```rust
            let cooccurrences = remaining_interactions
	                .join_map(&remaining_interactions, |_user, &item_a, &item_b| (item_a, item_b))
```
join can arrange `remaining_interactions` once and re-use it as both of its inputs, whereas right now it will maintain the same data twice. If that is a non-trivial collection (of all maintained data) that could be an improvement!

sscdotopen:
Hi Frank,

I should have provided a bit more context. This is actually meant to replace the `seb.rs` example, as it is an end-to-end implementation of the problem adressed there. I thought it my simply be interesting to have an example around that is a bit different from the usual graph algorithms.

I am trying to add the optimization you described but it seems  I cannot used an Arranged as a join input, can you give me a hint on how to proceed?

sscdotopen:
I tried
```
let arranged_remaining_interactions = remaining_interactions.arrange_by_key();

let cooccurrences = arranged_remaining_interactions
  .join_core(&arranged_remaining_interactions, |_user, &item_a, &item_b| (item_a, item_b))
  .filter(|&(item_a, item_b)| item_a > item_b)
  .count();
```
but that gives me the following error:

```
error[E0277]: `(_, _)` is not an iterator
  --> examples/itembased_cf.rs:42:18
   |
42 |                 .join_core(&arranged_remaining_interactions, |_user, &item_a, &item_b| (item_a, item_b))
   |                  ^^^^^^^^^ `(_, _)` is not an iterator
   |
   = help: the trait `std::iter::Iterator` is not implemented for `(_, _)`
   = note: required because of the requirements on the impl of `std::iter::IntoIterator` for `(_, _)`
```

frankmcsherry:
Yup. The `join_core` operator is super-flexible, and part of that is that the result selector closure should produce something that implements `IntoIterator`. If you wrap `(item_a, item_b)` as `Some((item_a, item_b))` you should be good to go (and it should compile away).

frankmcsherry:
And in fact, you could change the closure to be
```rust
|_user, &item_a, &item_b| {
    if item_a > item_b { Some((item_a, item_b)) }
    else { None }
}
```
and then you can skip the filter!

frankmcsherry:
Relatedly, this code
```rust
    let jaccard_similarities = cooccurrences
        .map(|((item_a, item_b), num_cooc)| (item_a, (item_b, num_cooc)))
        .join_map(&num_interactions_per_item, |&item_a, &(item_b, num_cooc), &occ_a| {
            (item_b, (item_a, num_cooc, occ_a))
        })
        .join_map(
            &num_interactions_per_item,
            |&item_b, &(item_a, num_cooc, occ_a), &occ_b| {
                ((item_a, item_b), (num_cooc, occ_a, occ_b))
            },
        )
        // We have to do this calculation in a map due to the lack of a total order for f64
        .map(|((item_a, item_b), (num_cooc, occ_a, occ_b))| {
            let jaccard = num_cooc as f64 / (occ_a + occ_b - num_cooc) as f64;
            ((item_a, item_b), jaccard)
        });
```
could be an arrangement of `num_interactions_per_item`, which then gets used twice. The final map can plausibly be done in the join, but I'm guessing your comment means that I have prevented that for some reason (probably: join tries to consolidate its results as it produces them, and `f64` not implementing `Ord` nor `Eq` means there is no sane logic to "accumulate" outputs).

sscdotopen:
Added the optimizations.

frankmcsherry:
Well, looks good to me! I bet there are a bunch of other performance hacks one could do, and if it comes to you wanting to assess the performance, let me know!

sscdotopen:
Happy to have contributed :) I actually have a COST baseline for this algorithm at https://github.com/schelterlabs/projects-amnesia/blob/master/src/itembased.rs I might want to do a performance comparison between these in the future, I'll ping you in case I start to work on that!

PR #200, opened 2019-08-14
Allow explicit frontier

frankmcsherry:
This PR is a minor edit to the new `import_frontier` method, which was great but which did not allow one to specify the frontier. For clear semantics, it can be important for an external agent to choose a consistent frontier across all workers, rather than just whatever their current compaction frontier happens to be.

cc: @comnik for sanity check.


comnik:
Uh this might be an important sanity check for other parts of 3DF as well. But I *think* we're good, because we bump advance frontiers consistently across workers. Definitely seems sane to have the ability.

PR #201, opened 2019-08-15
Log state drops

frankmcsherry:
This PR corrects what seems to have been an oversight, that we didn't log arranged state **drops**, so cumulative counts of maintained state would be wrong. Drops only happen when a dataflow is shut down or a trace handle dropped; they don't happen in the normal execution of a dataflow, so this shouldn't impact folks other than those who start and stop dataflows.

It is a breaking change, in that we have a new enum variant for exhaustive matches, but such is the cost of progress!

cc @comnik @ryzhyk

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -23,6 +23,8 @@ pub enum DifferentialEvent {
     Batch(BatchEvent),
     /// Merge start and stop events.
     Merge(MergeEvent),
+    /// Batch dropped when trace dropped.
+    Drop(DropEvent),
     /// A merge failed to complete in time.
     MergeShortfall(MergeShortfall),
 }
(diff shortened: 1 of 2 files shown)


comnik:
Uh nice, I ran into this recently, thought it was intended behaviour. Time for a `tdiag` update!

frankmcsherry:
I'm still not seeing the output I expect, so there might be some latent bugs. It looks like it should be logging things, but I'm not actually getting the events. More PRs may be inbound!

frankmcsherry:
Ah, the spine clears its batches without logging if it gets an advance by to `&[]`. Fix incoming.u

frankmcsherry:
Fix pushed at master accidentally. Oops. Let's see if it works. :D

Update: accidentally broke master. It's been a while; couldn't keep the streak going. Actual fix coming once tests land.

PR #213, opened 2019-09-04
Present

frankmcsherry:
This PR add a new difference type, `Present`, which is a zero-sized type indicating that the record is present. It supports settings like batch computation and some logic programs where facts are only ever introduced, and never retracted. The type has no zero, and addition and multiplication are the identity operations.

To use the type, it was important to generalize `ThresholdTotal`, in a way that I am not 100% satisfied with, but we can iterate on that.

cc @comnik


comnik:
Exciting!

PR #216, opened 2019-09-06
Log arrangement sharing

frankmcsherry:
This PR introduces logging statements in the construction, clone, and drop of the `TraceAgent` type, which logs the wrapped operator id and a signed integer indicating the change in occurrence count for the agent (plus one on construction and clone, minus one on drop).

This information is meant to allow introspection into the amount of re-use in the system, for diagnostic purposes (to see if sharing is effective, and to report the degree for celebration).

cc @comnik @utaal


utaal:
Woooo!

PR #220, opened 2019-10-04
Cross empty batches better

### LINKED ISSUE #219: Memory leak in differential
I believe we ran into a memory leak in DD master. It seems to happen when we haven't fed enough input data to activate some of the operators in the program, but keep incrementing the timestamp on input sessions. Here is a minimal failing example based on `examples/bfs.rs`. When I run it, I see the memory usage of the program grow unboundedly.

Note that uncommenting the line marked with
```
// !!! UNCOMMENTING THIS LINE ELIMINATES THE LEAK !!!`
```
eliminates the leak, with the memory usage remaining constant.

```
extern crate rand;
extern crate timely;
extern crate differential_dataflow;

use timely::dataflow::*;
use timely::dataflow::operators::probe::Handle;

use differential_dataflow::input::Input;
use differential_dataflow::Collection;
use differential_dataflow::operators::*;
use differential_dataflow::lattice::Lattice;
use differential_dataflow::logging::DifferentialEvent;

type Node = u32;
type Edge = (Node, Node);

fn main() {

    // define a new computational scope, in which to run BFS
    timely::execute_from_args(std::env::args(), move |worker| {

        if let Ok(addr) = ::std::env::var("DIFFERENTIAL_LOG_ADDR") {

            eprintln!("enabled DIFFERENTIAL logging to {}", addr);

            if let Ok(stream) = ::std::net::TcpStream::connect(&addr) {
                let writer = ::timely::dataflow::operators::capture::EventWriter::new(stream);
                let mut logger = ::timely::logging::BatchLogger::new(writer);
                worker.log_register().insert::<DifferentialEvent,_>("differential/arrange", move |time, data|
                    logger.publish_batch(time, data)
                );
            }
            else {
                panic!("Could not connect to differential log address: {:?}", addr);
            }
        }

        // define BFS dataflow; return handles to roots and edges inputs
        let mut probe = Handle::new();
        let (mut roots, mut graph) = worker.dataflow(|scope| {

            let (root_input, roots) = scope.new_collection();
            let (edge_input, graph) = scope.new_collection();

            let result = bfs(&graph, &roots);

            result.map(|(_,l)| l)
                  .consolidate()
                  .probe_with(&mut probe);

            (root_input, edge_input)
        });

        roots.insert(0);
        roots.close();

        // !!! UNCOMMENTING THIS LINE ELIMINATES THE LEAK !!!
        //graph.insert((0,1));
        graph.advance_to(1);
        graph.flush();

        loop {
            let time = *graph.time();
            graph.advance_to(time + 1);
            graph.flush();
            while probe.less_than(graph.time()) {
                worker.step_or_park(None);
            }
        }

    }).unwrap();
}

// returns pairs (n, s) indicating node n can be reached from a root in s steps.
fn bfs<G: Scope>(edges: &Collection<G, Edge>, roots: &Collection<G, Node>) -> Collection<G, (Node, u32)>
where G::Timestamp: Lattice+Ord {

    // initialize roots as reaching themselves at distance 0
    let nodes = roots.map(|x| (x, 0));

    // repeatedly update minimal distances each node can be reached from each root
    nodes.iterate(|inner| {

        let edges = edges.enter(&inner.scope());
        let nodes = nodes.enter(&inner.scope());

        inner.join_map(&edges, |_k,l,d| (*d, l+1))
             .concat(&nodes)
             .reduce(|_, s, t| t.push((*s[0].0, 1)))
     })
}
```

I realize this looks artificial, but we also observed this in more realistic scenarios, when it just happens so that that some inputs of a computation remain empty.  I don't believe the bug existed in 0.9, but I haven't bisected exactly when it was introduced.

@frankmcsherry 
@mbudiu-vmw
@ekcs

frankmcsherry:
This PR attempts to fix #219 which observes memory leaks if data are not inserted. What seems to have happened is that the logic for spanning empty batches erroneously started from the upper limit of batches, where it should have started from the lower limit of batches. This was everywhere correct except when the first batch is empty, in that we would fail to see it.

In addition, both `threshold_total` and `count_total` erroneously announced themselves as `unary` operators, which do not get progress notifications. This means that they would also stall on intervals of empty time. That has also been fixed.

cc @ryzhyk

### DIFF
--- a/src/operators/count.rs
+++ b/src/operators/count.rs
@@ -74,7 +74,7 @@ where
         let mut trace = self.trace.clone();
         let mut buffer = Vec::new();
 
-        self.stream.unary(Pipeline, "CountTotal", move |_,_| move |input, output| {
+        self.stream.unary_frontier(Pipeline, "CountTotal", move |_,_| move |input, output| {
 
             // tracks the upper limit of known-complete timestamps.
             let mut upper_limit = timely::progress::frontier::Antichain::from_elem(<G::Timestamp>::minimum());
--- a/src/operators/join.rs
+++ b/src/operators/join.rs
@@ -322,8 +322,9 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
         let mut trace2 = Some(other.trace.clone());
 
         // acknowledged frontier for each input.
-        let mut acknowledged1: Option<timely::progress::frontier::Antichain<G::Timestamp>> = None;
-        let mut acknowledged2: Option<timely::progress::frontier::Antichain<G::Timestamp>> = None;
+        use timely::progress::frontier::Antichain;
+        let mut acknowledged1: Option<Antichain<G::Timestamp>> = None;
+        let mut acknowledged2: Option<Antichain<G::Timestamp>> = None;
 
         // deferred work of batches from each input.
         let mut todo1 = Vec::new();
@@ -428,6 +429,10 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
                 if trace2.is_some() && input1.frontier().is_empty() { trace2 = None; }
                 if let Some(ref mut trace2) = trace2 {
                     trace2.advance_by(&input1.frontier().frontier()[..]);
+                    // At this point, if we haven't seen any input batches we should establish a frontier anyhow.
+                    if acknowledged2.is_none() {
+                        acknowledged2 = Some(Antichain::from_elem(<G::Timestamp>::minimum()));
+                    }
                     if let Some(acknowledged2) = &mut acknowledged2 {
                         trace2.advance_upper(acknowledged2);
                         trace2.distinguish_since(acknowledged2.elements());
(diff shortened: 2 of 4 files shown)


ryzhyk:
Many thanks for the fix! Do you know when the bugs were introduced? I am asking because I did not observe this behavior in 0.9.

frankmcsherry:
Probably around then. I don't recall which features landed when, but empty batches had some work done in the 0.9 to 0.10 transition. Resources got stuck less (previously, memory leak on all empty intervals) but apparently the fix had issues. It may just have been different settings where the leaks happened (but the definitely happened before).

ryzhyk:
I still see the exact same behavior (memory footprint steadily increasing) using the code in #219 with the `empty_batch` branch.

frankmcsherry:
Interesting. The smaller repro was fixed for me, but let me retry your later repro.

frankmcsherry:
Ah, I see the issue. It's another flavor of the same thing, but in `join`. Such a nuisance!

frankmcsherry:
This should be ready to try out again (your repro seems to work fine for me now).

ryzhyk:
Thanks! Let me try it on the DDlog-generated program where we observed this originally.

ryzhyk:
Yep, it all looks healthy now. Thanks a lot for the quick fix!

ryzhyk:
Do you foresee any problems with merging this into `continual_merge`? (I can do it in my fork, just checking if you know of a reason it wouldn't work).

PR #231, opened 2019-11-19
Do not drop batches.

frankmcsherry:
Previously, if the advance frontier for a trace reached the empty set, indicating that no further requirements exist for correct accumulation, the trace would drop all batches. This has some problematic interactions with `read_upper()` and `map_batches()` which cannot tell that things are broken. It seems that there is no urgency to drop the batches, as they will be dropped the instant the last trace handle is dropped. If they are not dropped this way, it is because some weirdo is holding on to a defunct trace handle. I'm sure we will find out that I am such a weirdo, but it seems like the correct fix in that case is to drop the handle rather than monkey around with the implementation.

This also updates the logging for `spine_fueled_neu.rs` to correctly record batch drops when the trace is dropped.

cc: @umanwizard

### DIFF
--- a/src/trace/implementations/spine_fueled.rs
+++ b/src/trace/implementations/spine_fueled.rs
@@ -199,9 +199,14 @@ where
     }
     fn advance_by(&mut self, frontier: &[T]) {
         self.advance_frontier = frontier.to_vec();
-        if self.advance_frontier.len() == 0 {
-            self.drop_batches();
-        }
+
+        // Commenting out for now; causes problems in `read_upper()`.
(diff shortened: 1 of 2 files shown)


frankmcsherry:
My hope is that there is zero impact on "idiomatic" uses, meaning those in which mere moments after one expresses that they do not plan on reading from a trace they also drop the trace. It is very believable that there are non-idiomatic uses, including ones that make total sense and reveal that I am ret-conning "idiomatic" to mean "make sense to me at the moment".

frankmcsherry:
Well, let's try it out and hear if there are any regressions.

Issue #218, opened 2019-10-01
`.inspect()`-ing a collection after the dataflow has been created.


ryzhyk:
I wonder if there is a way to attach an `inspect()` operator to a collection after the dataflow graph has been created. This would be helpful in situations where I do not know ahead of time which collections I will need to inspect, and I do not want to pay even the small price of attaching `inspect()` to all collections ahead of time.

Thanks!

frankmcsherry:
I think the main way to do this at the moment is to export arrangements for things you might want to examine. If you want to inspect any of them, you can interactively scan around using cursors, or you can build a dataflow with an `inspect` operator. The down side is that you need to manage things like advancing frontiers, which is a pain to remember to do.

Timely is not otherwise very fond of dynamically modifying the dataflow graph, even though we both understand that it would not actually be of much consequence. In e.g. DDlog, I think most named collections have associated `Variable` things, each of which has an arrangement behind it (the `Distinct` uses one); you could snag each of these, and allow interactive inspection of these collections (as long as you make sure to downgrade the capabilities of each).

ryzhyk:
> If you want to inspect any of them, you can interactively scan around using cursors

Thanks! Is there an example of scanning an arrangement using a cursor somewhere? If not, do you think  you could  point me to the relevant APIs?

ryzhyk:
Ok, I think I figured out the cursors API. Any objections if I submit a PR with an example?

frankmcsherry:
Go for it! In the past, `examples/arrangement.rs` has had such an example, but it is not obviously well maintained. Also, sorry about the latency; technically on vacation atm, so a bit slow (and, .. lots of other issues too!).

PR #236, opened 2019-11-25
Added an example based on the cursor API.

ryzhyk:
An example illustrating the use of the cursor API (see #218)

@frankmcsherry, I hope I got it right, but please double-check that I am using the API as expected and am not making any invalid assumptions.


frankmcsherry:
This looks good. I can imagine expanding it in the future to show off the intent of the cursor contents (e.g. that one should accumulate diffs by restriction to certain times), which I hope is not outside the intent. The raw usage is pretty handy, to show off what can be done without any magic.

PR #237, opened 2019-11-26
dogs/Cargo.toml: Use timely 0.11 in dogs^3.

ryzhyk:
Point `dogsdogsdogs/Cargo.toml` to timely v0.11 (from crates.io) so that
it's consistent with the main DD crate.


frankmcsherry:
Ah, sorry! I think the right fix to this is to point everyone at master. There are timely fixes to pick up. I'll push that asap.

frankmcsherry:
This should be updated in master now. The main repo and all subprojects should point at timely master, rather than crates.io.

PR #240, opened 2019-12-04
Fix typo

MarcoPolo:
A couple more typos. Thanks for making this book, I found it very helpful. Any other resources you'd recommend? I'm thinking of making a UI toolkit around this.


frankmcsherry:
Thanks! The use of `effect` is actually correct; as a verb it means "to put in to place". Everything is was totally incorrect though, so thanks!

Issue #242, opened 2020-01-07
Compaction and trace wrappers do not compose.


frankmcsherry:
As discussed in #234, the combination of timestamp manipulating trace wrappers and compaction interact badly, in that a trace wrapper may inaccurately summarize frontiers back to the wrapped trace, resulting in the re-wrapped timestamps being farther forward than correctness allows.

In the particular example there, the `AltNeu` wrapper, which extends timestamps with a bit indicating "old" or "new", takes a frontier of `AltNeu::alt(2)` and communicates upwards that this means that the frontier it cares about is now `[2]`, which is not correct as one can still distinguish `AltNeu::neu(1)` from `AltNeu::neu(2)`.

A bit of light thinking suggests that this is not a problem for the product order wrapper, but some more formalism is clearly needed here.

Issue #234, opened 2019-11-25
Possible bug in dogs^3


ryzhyk:
I ran into a potential bug in dogs^3, where the output of a program changes with the number of workers:

The repro is here, including a detailed description in the commit message:
https://github.com/ryzhyk/differential-dataflow/tree/dogs_bug

ryzhyk:
@frankmcsherry, just wondering if you still plan to look into this at some point. There is no rush, but delta queries are really nice. We'd love to be able to eventually use them in DDlog, and I bet many other DD users feel the same way :)

Thanks,
Leonid

frankmcsherry:
Investigating! Thank you for the easy reproduction!

frankmcsherry:
I have a PR up that requires arrangement's `enter_at` method to take a timestamp "inverting" method that walks back timestamps to those of the outer scope, which is used to communicate frontier information back up. It would be neat for this to be automatic, but I don't see a way for that to work at the moment.

In the case of `validedges.rs` the change would be something like
```rust
let nodes_arr_alt = nodes_arr.enter_at(inner, |_,_,t|AltNeu::alt(t.clone()), |t| t.time.saturating_sub(1));
let nodes_arr_neu = nodes_arr.enter_at(inner, |_,_,t|AltNeu::neu(t.clone()), |t| t.time.saturating_sub(1));

let edges_arr1_alt = edges_arr1.enter_at(inner, |_,_,t|AltNeu::alt(t.clone()), |t| t.time.saturating_sub(1));
let edges_arr2_alt = edges_arr2.enter_at(inner, |_,_,t|AltNeu::alt(t.clone()), |t| t.time.saturating_sub(1));
```
whose additional method says that we should try and subtract one from each time, unless we are at the minimum element in which case no worries (and don't wrap please).

With this fix, the specific erroneous behavior vanishes for me (I think; a further reduced error case vanishes). It is definitely a bug, and this definitely fixes a bug, but there could be new ones now. :)

ryzhyk:
Thanks for the fix!  I will test the PR with the DDlog branch that knows how to use dogs^3 using several of our datasets and will report back on how that goes.

Do you know by any chance if #235 is a separate problem or another manifestation of this?

frankmcsherry:
No clue yet.

ryzhyk:
Results so far: 

- DDlog delta query tests that used to fail are now passing!! More testing is in progress.
[...]

PR #243, opened 2020-01-07
Require arrangement sharing to communicate frontier changes.

frankmcsherry:
This PR requires that an arrangement's `enter_at` method takes a method to invert timestamps, so that frontier statements about the wrapper trace can be communicated back to the wrapped trace.
This method requires some care, but for the moment you probably shouldn't be using these methods without knowing about what is going on here; we could hope for a future interface that is clearer and less error prone.

### DIFF
--- a/dogsdogsdogs/examples/delta_query.rs
+++ b/dogsdogsdogs/examples/delta_query.rs
@@ -56,15 +56,15 @@ fn main() {
                 let changes = edges.enter(inner);
 
                 // Each relation we'll need.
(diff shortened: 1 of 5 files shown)


frankmcsherry:
This fixes #242 and #234

ryzhyk:
It looks like f8f80fa0afb915ef7d932f6afef71e343cd0cbae broke dogs^3, which still expects `minimum` to be a member of `Lattice`, so `cargo build` fails in `dogsdogsdogs`.

frankmcsherry:
Weird. I guess I failed to cargo update or something, not sure what (things were working locally on commit). It now seems to build for me (corrected examples too), but my confidence is shaken!

Issue #247, opened 2020-02-07
Wrong `since` frontier for merged batch?


LorenzSelv:
https://github.com/TimelyDataflow/differential-dataflow/blob/e50c9ce938c9783d7a5eb7fdf25f047f2c6ed145/src/trace/implementations/ord.rs#L227

Consider the full set of updates of e.g. a simple wordcount dataflow (single-dimensional timestamps):
```
full = { ("hello", (), t=5, +1) }
```

Suppose we seal two batches: `batch1` contains the only update, `batch2` is empty.
```
batch1 = Description { lower: 0,  upper: 10, since: 0 }
         Content { ("hello", (), t=5, +1) }
batch2 = Description { lower: 10, upper: 20, since: 10 }
         Content { }
```

Suppose the times inside `batch1` are advanced to timestamp `15`
```
batch1 = Description { lower: 0,  upper: 10, since: 15 }
         Content { ("hello", (), t=15, +1) }  // times in this batch has been advanced to t=15
batch2 = Description { lower: 10, upper: 20, since: 10 }
         Content { }
```

Suppose we merge the two batches. According to the code, the resulting batch would have the following description / content
```
Description { lower: 0, upper: 20, since: 10 }
Content { ("hello", (), t=15, +1) }
```
The `since` frontier is taken to be the **minimum** of the two `since` frontiers of the batches, so `batch2.since = 10` is taken.

However, the [property of the batch](https://github.com/TimelyDataflow/differential-dataflow/blob/56a7eeddbbccce46c2132620c6dbb2f0bff7bcc7/src/trace/description.rs#L33) does not hold anymore.

Consider for example `t=12`, for which it should hold as it is larger than the `since` frontier.
There are no accumulated updates before such time, which is not the case if we consider the
original set of updates (`full`).

To conclude, I think the `since` frontier of the merged batch should be the **maximum** of the two original batches (`t=15` in this case). And it's easy to see that the property would hold for the case above.

The intuition that justifies this is that if a batch looses the capability to distinguish since a certain timestamp (as `batch1` did) also a resulting merged batch will not have such capability.

-------

On a similar note, I believe there is another issue in the `work` function that performs the merging.
https://github.com/TimelyDataflow/differential-dataflow/blob/e50c9ce938c9783d7a5eb7fdf25f047f2c6ed145/src/trace/implementations/ord.rs#L293

[...]

PR #252, opened 2020-02-16
Since frontier

LorenzSelv:
#247 

I looked for all occurrences of batch formation and used the minimum frontier as the since frontier as you suggested, hope I didn't miss any.

### DIFF
--- a/src/trace/implementations/ord.rs
+++ b/src/trace/implementations/ord.rs
@@ -14,7 +14,7 @@ use std::marker::PhantomData;
 use std::fmt::Debug;
 
 use ::difference::Semigroup;
-use lattice::Lattice;
+use lattice::{Lattice, antichain_join};
 
 use trace::layers::{Trie, TupleBuilder};
 use trace::layers::Builder as TrieBuilder;
@@ -224,14 +224,9 @@ where
 
 		assert!(batch1.upper() == batch2.lower());
 
-		let since = if batch1.description().since().iter().all(|t1| batch2.description().since().iter().any(|t2| t2.less_equal(t1))) {
-			batch2.description().since()
-		}
-		else {
-			batch1.description().since()
-		};
+		let since = antichain_join(batch1.description().since(), batch2.description().since());
 
-		let description = Description::new(batch1.lower(), batch2.upper(), since);
+		let description = Description::new(batch1.lower(), batch2.upper(), since.elements());
 
 		OrdValMerger {
 			lower1: 0,
@@ -291,7 +286,9 @@ where
 
 		// if we are supplied a frontier, we should compact.
 		if let Some(frontier) = frontier.as_ref() {
-			OrdValBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos)
+			OrdValBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos);
+			let new_since = antichain_join(self.description.since(), &frontier[..]);
+			self.description = Description::new(self.description.lower(), self.description.upper(), new_since.elements());
 		}
 
         *fuel -= effort;
@@ -543,14 +540,9 @@ where
 
 		assert!(batch1.upper() == batch2.lower());
 
-		let since = if batch1.description().since().iter().all(|t1| batch2.description().since().iter().any(|t2| t2.less_equal(t1))) {
-			batch2.description().since()
-		}
-		else {
-			batch1.description().since()
-		};
+		let since = antichain_join(batch1.description().since(), batch2.description().since());
 
-		let description = Description::new(batch1.lower(), batch2.upper(), since);
+		let description = Description::new(batch1.lower(), batch2.upper(), since.elements());
 
 		OrdKeyMerger {
 			lower1: 0,
@@ -616,6 +608,8 @@ where
 
 		if let Some(frontier) = frontier.as_ref() {
 			OrdKeyBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos);
+			let new_since = antichain_join(self.description.since(), &frontier[..]);
+			self.description = Description::new(self.description.lower(), self.description.upper(), new_since.elements());
 		}
 
         *fuel -= effort;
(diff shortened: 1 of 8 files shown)


src/trace/implementations/ord.rs
```diff
@@ -291,7 +286,9 @@ where
 
 		// if we are supplied a frontier, we should compact.
 		if let Some(frontier) = frontier.as_ref() {
-			OrdValBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos)
+			OrdValBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos);
+			let new_since = antichain_join(self.description.since(), &frontier[..]);
+			self.description = Description::new(self.description.lower(), self.description.upper(), new_since.elements());
```
frankmcsherry: I *think* this is redundant with the changes you've made just above, around lines 229. The `self.description` and `frontier` are going to be static across all of these calls, so that determination at the beginning of the merge set-up should hold for all of the calls as well. But, I could be wrong! Did you add this out of caution, or because there is a specific code path that requires it?

src/trace/implementations/ord.rs
```diff
@@ -616,6 +608,8 @@ where
 
 		if let Some(frontier) = frontier.as_ref() {
 			OrdKeyBatch::advance_builder_from(&mut self.result, frontier, initial_key_pos);
+			let new_since = antichain_join(self.description.since(), &frontier[..]);
+			self.description = Description::new(self.description.lower(), self.description.upper(), new_since.elements());
```
frankmcsherry: Possibly also redundant, this time with the lines at 545, but for the same reason as further above.

frankmcsherry:
This looks great to me! There are two places that I think might be redundant, because you already set up what I *think* is the right description before starting, and the fields that drive the new `since` stay static while a batch is being merged. But, it could be that there is an important reason to have them there that you've seen, so let me know what you think about the comments!

Thanks very much, though!

src/trace/implementations/ord.rs
frankmcsherry: Thinking about it now, the fact that `frontier` is static is only because of how the code is used, and it is definitely "more safe" to re-check each time things get called. It is a bit too bad, because 1. we "know" that it won't be a problem, and 2. it is some non-trivial effort to re-determine the since frontier. That's a totally reasonable reason to have this code, though, because in the future someone might mis-use these functions and call with varying frontiers.

frankmcsherry: One option: we could make `frontier` into a `bool` (weird) and use `self.description.since` for the content of `frontier` when the bool is set. I think in the current code we always set `frontier`, but the flexibility exists to merge without compacting. However, I'm not sure there is any value in compacting to anything less than the since frontier of the resulting batch. Maybe we should make that mandatory?

frankmcsherry: (( If we like that idea, ripping out the `frontier` stuff is something either of us can do in a follow-on PR; don't sweat that at the moment, but I am interested to hear your opinion. ))

LorenzSelv: https://github.com/TimelyDataflow/differential-dataflow/blob/ff2aeff062078f0a437f5bcafe3bf8f61bb6fb13/src/trace/implementations/spine_fueled_neu.rs#L855

If I'm not mistaken, this is the only occurrence in which the `begin_merge` function is used.

The "constructor" of the `Merger` takes only the two batches, and the resulting `since` frontier is computed using the `since` frontiers of the two batches. The compaction frontier, however, is not passed to the constructor, but instead is stored inside the `MergeVariant::InProgress` enum to be then passed to the `work` function.

Since the compaction frontier might be greater than the `since` frontier computed in the constructor, we need to update it.

As you correctly pointed out, given how the `work` function is called, the compaction frontier is fixed. 

I think the following refactoring would make it:

* pass the (optional) compaction frontier to the constructor (so we can determine the final `since` frontier right away), rather than to the `work` function.

* if a compaction frontier was passed, the `Merger` sets an internal flag to tell whether or not it should compact when the `work` function is called (the compaction would be performed using the previously computed `since` frontier).

* `MergeVariant::InProgress` variant would then need to store only the two batches being merged and the merger object (i.e. we don't store the compaction frontier anymore).

frankmcsherry: I think this is right, with the caveat that the third bullet probably wants to be "only the two batches being merged and the merger object and the internal flag to tell whether or not it should compact". 

Actually, I'm not certain why the current `frontier: Option<Vec<T>>` isn't part of the merger ... which would also address the problem (prevent using different frontiers as the merge proceeds). But I like you proposal most! We should use the `since` frontier, and have a flag to indicate if compaction should happen (e.g. in case we know that the frontier is below the lower bounds, and would have no effect).

LorenzSelv: I think it's better to store the flag inside the merger, and we set the flag to true if we were provided a compaction frontier on construction. I added a commit with the refactoring.

frankmcsherry: Aaaa, sorry. I didn't notice the last sentence.

src/trace/implementations/ord.rs
```diff
@@ -220,18 +221,20 @@ where
...
-		}
-		else {
-			batch1.description().since()
+		let since = {
+			let since = antichain_join(batch1.description().since(), batch2.description().since());
+			if let Some(compaction_frontier) = compaction_frontier {
+				antichain_join(since.elements(), compaction_frontier)
+			} else {
+				since
+			}
```
frankmcsherry: I wonder if this looks better as
```rust
let mut since = antichain_join(batch1.description().since(), batch2.description().since());
if let Some(compaction_frontier) = compaction_frontier {
   since = antichain_join(since.elements(), compaction_frontier);
}
```
I'm happy to clean that up in post, as long as you don't mind!

LorenzSelv: yeah it definitely does, feel free to refactor

PR #256, opened 2020-03-01
Move graph_map to a dev dependency in dogsdogsdogs

benesch:
This isn't actually used by dogsdogsdogs proper.


frankmcsherry:
Looks great!

PR #257, opened 2020-03-05
Streamline empty batch handling

### LINKED ISSUE #253: Surprising transient crash
Materialize had an apparently transient crash where differential reported
```
assertion failed: batch2.upper().iter().all(|t| acknowledged2.less_equal(t))
```
This is an assertion in `join` that ensures that newly received batches "properly" line up with acknowledged frontiers. Frontiers are acknowledged when they are received as the `upper` of a batch, or when we fast-forward over them because we have seen an empty batch whose `lower` matches the previously acknowledged frontier.

This doesn't reproduce, but it also isn't remotely clear what triggered it, so we should 1. watch out for these, and 2. try and simplify the logic around this code. To the second point, the code is complicated because of timely's requirement that messages contain a capability, which if removed would allow us to send empty batches along channels and remove the fast-forwarding logic.

To attempt to simplify debugging, the code was recently modified not only to panic, but also to print out the offending frontiers.

cc @benesch

frankmcsherry:
The various `import` operators on traces would produce empty batches, because they held a capability to do so. This confused some other operators that would fast-forward their acknowledged frontiers across empty regions (using information from the trace) and then apparently receive those empty batches. There is no real error, in that the region of time was already confirmed empty, and the received empty batches do not indicate actual work to do.

fixes #253

### DIFF
--- a/src/operators/arrange/agent.rs
+++ b/src/operators/arrange/agent.rs
@@ -322,8 +322,10 @@ where
                                 },
                                 TraceReplayInstruction::Batch(batch, hint) => {
                                     if let Some(time) = hint {
-                                        let delayed = capabilities.delayed(&time);
-                                        output.session(&delayed).give(batch);
+                                        if !batch.is_empty() {
+                                            let delayed = capabilities.delayed(&time);
+                                            output.session(&delayed).give(batch);
+                                        }
                                     }
                                 }
                             }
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Going to land this to attempt to exercise it on real workloads. I think it should be mostly harmless, unless suppressing the empty batches coming out of the `import` operators somehow confuses the system in a different way.

PR #263, opened 2020-03-24
Upsert support

frankmcsherry:
This PR introduces a prototype "upsert" arrangement operator. Instead of ingesting standard update triples, it receives keyed optional values, where the current value for a key is the most recently received value. 

This is substantially less general than differential dataflow allows, but lots of people use this format. The reason to do this here is that if the plan is to arrange the results anyhow, then we can re-use the arrangement and extract the "prior" values associated with keys, and undo the damage done at nominal incremental overhead.

cc @wangandi


frankmcsherry:
The code is mostly a copy/paste of the `Arrange` operator, and much of its generality isn't required here, where a total order on timestamps is required. That should probably be cleaned up, but it seemed most likely to work correctly starting from something else that was known (believed) to be correct.

rjnn:
🎉!

PR #264, opened 2020-03-28
Properly consolidate slices with one element

benesch:
If that one element had a multiplicity of zero, consolidate would
improperly fail to discard it.


frankmcsherry:
I believe this is behavior that differential doesn't want to commit to. We can certainly make the change for now, but I'll also fix the docs to clarify the intent (and we should fix the use in Materialize). Differential doesn't produce zero diffs as inputs for these methods (if it does, that is a bug) and as these methods live at the computational heart of what it does, any unnecessary requirements on them are undesirable.

PR #265, opened 2020-03-28
Update documentation and simplify implementation

frankmcsherry:
The documentation for `consolidate` resulted in surprises elsewhere when the method did not fulfill a very natural expectation (that it would filter supplied zero diff elements). The behavior was improved in #264, but going forward it would be best to manage the expectations about these methods.

cc @benesch


benesch:
(Maybe "correct" is too strong a word for your taste, but you know what I mean.)

PR #267, opened 2020-04-02
Update mdbook to Timely 0.11.1 and Differential 0.11.0

sugohugo:
So, I'm not sure if this is useful at all, but I thought it might be worth a try. When following the examples from the `mdbook` I ran into some errors - mainly because I skipped the very first chapter where it said that one should use differential v0.7. 
Apparently, some of the output formatting changed and most importantly the result from the `join()` operator. This lead to quite a bit of confusion on my side because I tried to debug something locally using the `mdbook` as documentation and taking it at face value. 
I also adapted the examples in `chapter_a` a bit because it was not possible to simply replace code parts from the previous example with the snippets without declaring some variables in the scope at first (might be difficult to see for a beginner in differential).

Maybe this PR can save someone else the struggle in the future, and I hope that I managed to update it consistently.

### DIFF
--- a/mdbook/src/chapter_0/chapter_0_0.md
+++ b/mdbook/src/chapter_0/chapter_0_0.md
@@ -21,8 +21,8 @@ Instead, edit your `Cargo.toml` file, which tells Rust about your dependencies, 
         authors = ["Your Name <your_name@you.ch>"]
 
         [dependencies]
-        timely = "0.7"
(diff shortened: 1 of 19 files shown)


frankmcsherry:
These all look great! Thanks very much. I was literally looking at the `0.7` stuff last night thinking "I should update that tomorrow"... Sorry!

PR #268, opened 2020-04-18
Update README.md

owenrh:
Code example was incorrect - has semi-colon instead of a dot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ For example, here is a differential dataflow fragment to compute the out-degree 
 ```rust
 let out_degr_dist =
 edges.map(|(src, _dst)| src)    // extract source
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Ah, wait! You are absolutely correct that the semicolon is wrong, but the dot is already on the next line! The right fix is just to drop the semicolon, I believe.

frankmcsherry:
Looks great!

owenrh:
Hmm, actually the code snippet is in the README twice, I only fixed it in one place. One secondo.

owenrh:
Patched the second example as well.

frankmcsherry:
Looks great; thanks very much for the attention to detail!

PR #281, opened 2020-09-10
Generalize arrangements to Containers

frankmcsherry:
This PR-in-progress generalizes the storage used for arrangement batches to a `Container` trait that informally looks like a `Vec` but is not obliged to have that implementation. In particular, we are interested in the `Columnation` implementation, which maintains a `Vec`-like storage but whose items can arrange their owned members in contiguous memory.

### DIFF
--- a/src/trace/implementations/ord.rs
+++ b/src/trace/implementations/ord.rs
@@ -66,7 +66,6 @@ impl<T> RetainFrom<T> for Vec<T> {
         let mut write_position = index;
         for position in index .. self.len() {
             if predicate(position, &self[position]) {
-                // TODO: compact the inner region and update pointers.
                 self.swap(position, write_position);
                 write_position += 1;
             }
@@ -212,7 +211,7 @@ where
                 keys_pos += 1;
             }
             let lower = vals_off[index].try_into().unwrap();
-            let upper =    vals_off[index+1].try_into().unwrap();
+            let upper = vals_off[index+1].try_into().unwrap();
             if lower < upper {
                 vals_off[write_position+1] = vals_off[index+1];
                 write_position += 1;
@@ -228,8 +227,8 @@ where
         let offs = &mut layer.offs;
         let mut write_position = key_start;
         layer.keys.retain_from(key_start, |index, _item| {
-            let lower =    offs[index].try_into().unwrap();
-            let upper =    offs[index+1].try_into().unwrap();
+            let lower = offs[index].try_into().unwrap();
+            let upper = offs[index+1].try_into().unwrap();
             if lower < upper {
                 offs[write_position+1] = offs[index+1];
                 write_position += 1;
@@ -575,8 +574,8 @@ where
         let offs = &mut layer.offs;
         let mut write_position = key_start;
         layer.keys.retain_from(key_start, |index, _item| {
-            let lower =    offs[index].try_into().unwrap();
-            let upper =    offs[index+1].try_into().unwrap();
+            let lower = offs[index].try_into().unwrap();
+            let upper = offs[index+1].try_into().unwrap();
             if lower < upper {
                 offs[write_position+1] = offs[index+1];
                 write_position += 1;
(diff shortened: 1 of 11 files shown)


src/trace/layers/mod.rs
```diff
@@ -106,6 +109,70 @@ pub trait Cursor<Storage> {
...
+        }
+    }
+    fn with_capacity(_size: usize) -> Self {
+        Self::default()
+    }
+    fn reserve(&mut self, _additional: usize) {
+    }
+    fn merge_capacity(_cont1: &Self, _cont2: &Self) -> Self {
+        Self::default()
+    }
```
frankmcsherry, line 172: These look like they should be implemented in order to be competitive.

src/trace/implementations/ord.rs
```diff
@@ -46,67 +49,112 @@ pub type OrdKeySpine<K, T, R, O=usize> = Spine<Rc<OrdKeyBatch<K, T, R, O>>>;
...
+    /// Retains elements from an index onwards that satisfy a predicate.
+    fn retain_from<P: FnMut(usize, &T)->bool>(&mut self, index: usize, predicate: P);
+}
+
+impl<T> RetainFrom<T> for Vec<T> {
+    fn retain_from<P: FnMut(usize, &T)->bool>(&mut self, index: usize, mut predicate: P) {
+        let mut write_position = index;
+        for position in index .. self.len() {
+            if predicate(position, &self[position]) {
+                // TODO: compact the inner region and update pointers.
```
frankmcsherry, line 69: What does this mean? :D

src/trace/implementations/ord.rs
```diff
@@ -151,62 +199,59 @@ where
...
+            // As we pass each key offset, record its new position.
+            if index == keys_off[keys_pos].try_into().unwrap() {
+                keys_off[keys_pos] = O::try_from(write_position).unwrap();
+                keys_pos += 1;
             }
-            // batch.layer.offs[i+1] = write_position;
-        }
-        layer.vals.keys.truncate(write_position);
+            let lower = vals_off[index].try_into().unwrap();
+            let upper =    vals_off[index+1].try_into().unwrap();
```
frankmcsherry, line 215: Tab that should be a space, looks like.

src/trace/implementations/ord.rs
```diff
@@ -151,62 +199,59 @@ where
...
+        let offs = &mut layer.offs;
         let mut write_position = key_start;
-        for i in key_start .. layer.keys.len() {
-
-            let lower: usize = layer.offs[i].try_into().unwrap();
-            let upper: usize = layer.offs[i+1].try_into().unwrap();
-
+        layer.keys.retain_from(key_start, |index, _item| {
+            let lower =    offs[index].try_into().unwrap();
+            let upper =    offs[index+1].try_into().unwrap();
```
frankmcsherry, line 232: Two tabs that should be two spaces, looks like.

src/trace/implementations/ord.rs
frankmcsherry: Looks vestigial, as it also appears (with clearer meaning) in the `columnation` code.

src/trace/layers/mod.rs
```diff
@@ -106,6 +109,73 @@ pub trait Cursor<Storage> {
...
+        self.reserve_items(slice.iter());
+        for item in slice.iter() {
+            self.copy(item);
+        }
+    }
+    fn with_capacity(size: usize) -> Self {
+        Self::with_capacity(size)
+    }
+    fn reserve(&mut self, _additional: usize) {
+    }
```
frankmcsherry, line 170: We use this in various `push_merge` implementations for trie layers. I don't think it should block merging this PR, but probably worth pondering at least briefly. For example:
```rust
        let (trie1, mut lower1, upper1) = other1;
        let (trie2, mut lower2, upper2) = other2;

        self.vals.reserve((upper1 - lower1) + (upper2 - lower2));
```

PR #287, opened 2020-11-10
Add `histogram` method

davidspies:
I'm finding in general that I consistently have to add special-case logic to get 0-counts to show up when calling `count`. This method would make that unnecessary.

### DIFF
--- a/src/operators/mod.rs
+++ b/src/operators/mod.rs
@@ -4,7 +4,7 @@
 //! operators have specialized implementations to make them work efficiently, and are in addition
 //! to several operations defined directly on the `Collection` type (e.g. `map` and `filter`).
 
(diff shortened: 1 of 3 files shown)


frankmcsherry:
If you'd like there is another way to do this that is less overhead, in case you don't expect to have negative counts:

```
self.concat(keys)
    .count()
    .map(|(key, count)| (key, count-1));
```

I'm hesitant to put helper methods in to the core DD library, especially when there could be trade-offs in terms of which implementation folks might prefer (e.g. do you need sensitivity to negative counts or not). My understanding (and complain if I am wrong) is that the same code in your codebase would remove mitigate the issue of recurring special-cased logic.

davidspies:
That makes sense

PR #288, opened 2020-11-10
Function for concatenating rather than replacing

davidspies:
In the case where the variable is being accumulated by concatenating on a shrinking diff, it may be more efficient to skip the negate-and-then-un-negate part.

### DIFF
--- a/src/operators/iterate.rs
+++ b/src/operators/iterate.rs
@@ -188,12 +188,17 @@ impl<G: Scope, D: Data, R: Abelian> Variable<G, D, R> where G::Timestamp: Lattic
 
     /// Adds a new source of data to the `Variable`.
     pub fn set(self, result: &Collection<G, D, R>) -> Collection<G, D, R> {
+        let in_result = self.source.negate().concat(result);
+        self.set_concat(&in_result)
+    }
+
+
+    /// Concat a new source of data to the data already in the `Variable`.
+    pub fn set_concat(self, result: &Collection<G, D, R>) -> Collection<G, D, R> {
         let step = self.step;
-        self.source
-            .negate()
-            .concat(result)
+        result
             .inner
-            .flat_map(move |(x,t,d)| step.results_in(&t).map(|t| (x,t,d)))
+            .flat_map(move |(x, t, d)| step.results_in(&t).map(|t| (x, t, d)))
             .connect_loop(self.feedback);
 
         self.collection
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Sorry for the delay.

Talking this out, this looks like the `SemigroupVariable`, but is perhaps appropriate for folks who don't know statically that they will have the structural guarantee it provides. If that is the intent, it would be great to put some comments in to that effect (the comments in the file could generally be improved, which is on me). In particular, it would be great to say what `set_concat` corresponds to (`self.set(self.concat(result)`, I think) and why one would use it (performance).

src/operators/iterate.rs
```diff
@@ -188,12 +188,17 @@ impl<G: Scope, D: Data, R: Abelian> Variable<G, D, R> where G::Timestamp: Lattic
...
+    /// Concat a new source of data to the data already in the `Variable`.
+    pub fn set_concat(self, result: &Collection<G, D, R>) -> Collection<G, D, R> {
         let step = self.step;
-        self.source
-            .negate()
-            .concat(result)
+        result
             .inner
-            .flat_map(move |(x,t,d)| step.results_in(&t).map(|t| (x,t,d)))
+            .flat_map(move |(x, t, d)| step.results_in(&t).map(|t| (x, t, d)))
```
frankmcsherry, line 201: The spacing change here seems to be noise, and makes the code different from `SemigroupVariable` which would otherwise be identical to the text here, I think.

davidspies: I didn't notice `SemigroupVariable`, but `SemigroupVariable` seems to be undocumented (or rather the documentation was copy-pasted from `Variable` without any changes) and also doesn't support `new_from` and I don't see any way to replicate the behavior of `new_from` since `set` _consumes_ the variable.
I would advocate for deleting (or deprecating for backwards compatibility) `SemigroupVariable` in favor of this method.

frankmcsherry: Unfortunately, `SemigroupVariable` is crucial for lots of use cases. In particular, it has weaker requirements on the diff field that allow it to be use with types that implement `Semigroup` rather than `Abelian` (which is strictly more restrictive than `Semigroup`).

frankmcsherry: It seems reasonable to have both, at least the case where folks can't statically determine that they can use `SemigroupVariable` it would still be nice to produce the simpler dataflow that it constructs. I'm ok with the idea that the PR lands as is, but I'll probably drop in and write a bunch of comments to clarify things (you are right that `SemigroupVariable`'s docs are nothing to be proud of).

frankmcsherry:
Going to merge a fast-follow with some refreshed documentation (re-reading it myself it now seems very problematic).

PR #292, opened 2020-12-27
operators/join.rs: fix documentation for `join`

pkhuong:
The copy/paste from join_map was a bit too eager.

### DIFF
--- a/src/operators/join.rs
+++ b/src/operators/join.rs
@@ -28,7 +28,9 @@ use trace::TraceReader;
 /// Join implementations for `(key,val)` data.
 pub trait Join<G: Scope, K: Data, V: Data, R: Semigroup> {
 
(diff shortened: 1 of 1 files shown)


pkhuong:
While this minimal doc patch is a simple improvement, might it make sense to point readers at `join_map`, in (the likely) case that's a more direct way to express what they want?

frankmcsherry:
That makes sense! `join` almost never leaves the data in the form that the next operator can best use it.

pkhuong:
The new internal link uses https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html (stabilised in 1.48).

Render: Screenshot 2020-12-27 at 18 04 02

frankmcsherry:
Looks great; thank you very much!

PR #295, opened 2021-01-11
Track timely Config changes

frankmcsherry:
This PR tracks changes to timely's `Configuration` type, which is now `Config` and handles a broader class of options than before. In particular, it houses user-defined configurations, which we are attempting with this PR. Rather than specify `DIFFERENTIAL_EAGER_MERGE` in the environment, a user can add an `isize` value with key `differential/idle_merge_effort` to `worker.config()` which is then read out by the worker at arrangement construction time. This is a similar idiom to user-specific logging channels, which seems to have worked well enough.

The main downside here is that it was super casual to just screw around with your environment. Now, interested programs must solicit this value somehow, and explicitly set it as part of the configuration. That is more work (and in particular isn't magically done for existing DD programs). This is I suppose the downside of DD not having an execution prelude in the same way that timely has the `execute` method.

cc: @benesch, @ryzhyk

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -25,11 +25,11 @@ pub trait Input : TimelyInput {
     /// extern crate timely;
     /// extern crate differential_dataflow;
     ///
-    /// use timely::Configuration;
+    /// use timely::Config;
     /// use differential_dataflow::input::Input;
     ///
     /// fn main() {
-    ///     ::timely::execute(Configuration::Thread, |worker| {
+    ///     ::timely::execute(Config::thread(), |worker| {
     ///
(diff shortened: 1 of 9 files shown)


benesch:
This seems like a great start to me! One idea I had for a halfway point between `differential::execute` and this was to add a method like `differential::configure` that worked like so:

```rust
let config = timely::Config::from_args(std::env::args());
differential::config(config, differential::Config { eager_merge: 1000 });
timely::execute(config, ...)
```

Or via an extension trait:

```rust
use differential::Config;
let config = timely::Config::from_args(std::env::args()).differential_eager_merge(1000);
timely::execute(config, ...)
```

PR #297, opened 2021-01-25
Disable timely's default features

benesch:
Allowing disabling timely's default features (i.e., getopts), though they are still enabled by default.


frankmcsherry:
Looks good; will merge once green (which I expect)! Thanks very much.

PR #300, opened 2021-01-28
Generalize arrange to change type of records

frankmcsherry:
This PR introduces a generalization to the `arrange` method, which now allows the user to interpose between the batch formation and accepting the batch into the output trace. In particular, the input and output trace formats may be different, which allows the user the ability to perform some non-standard translation, for example playing a state machine forward and recording the transitions in the output.

The vanilla `arrange_core` operator is now implemented using "logic" that just passes through the input batch. It looks like so:
```rust
        self.arrange_general::<P, Tr, Tr, _, _>(pact, name, |_capability, _trace_agent| {
            |batch, _capability| (batch, Vec::new())
        })
```
Other operators like `upsert` and `reduce` could in principle be ported to this framework, but I wanted to float this first before doing a massive re-write.

cc: @ruchirK @petrosagg

### DIFF
--- a/src/operators/arrange/arrangement.rs
+++ b/src/operators/arrange/arrangement.rs
@@ -517,29 +517,62 @@ where
         Tr::Batch: Batch<K, V, G::Timestamp, R>,
         Tr::Cursor: Cursor<K, V, G::Timestamp, R>,
     {
-        // The `Arrange` operator is tasked with reacting to an advancing input
-        // frontier by producing the sequence of batches whose lower and upper
-        // bounds are those frontiers, containing updates at times greater or
-        // equal to lower and not greater or equal to upper.
-        //
-        // The operator uses its batch type's `Batcher`, which accepts update
-        // triples and responds to requests to "seal" batches (presented as new
-        // upper frontiers).
-        //
-        // Each sealed batch is presented to the trace, and if at all possible
-        // transmitted along the outgoing channel. Empty batches may not have
-        // a corresponding capability, as they are only retained for actual data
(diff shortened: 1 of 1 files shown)


src/operators/arrange/arrangement.rs
```diff
@@ -517,29 +517,62 @@ where
...
+    G: Scope,
+    G::Timestamp: Lattice+Ord,
+    K: ExchangeData+Hashable,
+    V: ExchangeData,
+    R: Semigroup+ExchangeData,
+{
+    /// Arranges a differential dataflow collection with custom user logic.
+    ///
+    /// This method generalizes `arrange` in that the output type may differ
+    /// from the input type, and the user is allow to perform logic as the
```
ruchirK, line 538: nit: is allowed

ruchirK (approved):
LGTM

frankmcsherry:
Brief thoughts on `reduce` at least: this probably isn't a great place to port that too without some more thinking, as one of its goals is to accept pre-arranged input, and this operator is doing the arranging. But, I could imagine with a bit more thinking finding a way to blend the two, where you get a stream of arranged data and input trace, which instances of this operator would then immediately drop (just because that is what it does) but which others (like `reduce`) could hold on to.

It's a bit weird, because this operator would generalize to "something that takes arrangements as input" which .. well at the moment it is what makes arrangements, so clearly there is some unpicking to do there.

The goal, though, is to avoid having so many copy/paste instances of "operator that forms batches and maintains traces" as we have in `arrange`, `reduce`, and `upsert`.

PR #305, opened 2021-02-17
Work around apparent timely glitch

frankmcsherry:
I'm seeing a weird glitch out of timely dataflow in a heavily loaded scenario (32 worker threads on my laptop) that seems to provide incorrect frontier information to the `lookup_map` operator. I'm investigating that independently, as the reach of that issue could be much larger, but this fix is sane anyhow: directly consult the trace to see what times it views as closed rather than making the leap from input frontier to that quantity indirectly.

### DIFF
--- a/dogsdogsdogs/src/operators/lookup_map.rs
+++ b/dogsdogsdogs/src/operators/lookup_map.rs
@@ -78,13 +78,16 @@ where
         // to observe the frontier and to drive scheduling.
         input2.for_each(|_, _| { });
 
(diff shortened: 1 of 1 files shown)


frankmcsherry:
I've tracked down and I believe fixed the timely glitch, over in https://github.com/TimelyDataflow/timely-dataflow/pull/360. I think we can close this, or merge it as looking directly at the trace frontier does seem smarter. Perhaps it can linger while we test out the timely fix.

PR #307, opened 2021-02-19
Robust join implementation

frankmcsherry:
This PR changes the implementation of `join` to be more robust in the presence of pre-compacted arrangements.

In particular, https://github.com/MaterializeInc/materialize/issues/5774 was at its core about `join` receiving pre-compacted arrangements, misunderstanding the physical compaction frontier, mis-*SETTING* the physical compaction frontier, and then believing it was getting results that it was not correctly receiving. A giant tire fire, really.

The problem was that it was not carefully tracking the relationship between received batches and the physical compaction frontier. This PR attempts to fix that by being substantially more clear about the invariants that should be checked at operator construction time, and then maintained as the operator executes. They are not very complicated, and boil down to:

> for each trace, the physical compaction frontier should be less or equal to the completed trace frontier.

That is, we shouldn't physically compact *past* the timestamps we have completed, which makes some sense because such a handle to a trace is not especially useful (you cannot subset out any of the existing data, under the current rules for extracting out subsets of data). Join in particular wants to be able to subset out from a trace the data that it has received in batches, and if it cannot reliably do that .. well the implementation we have now doesn't work, nor will this new one (just, it will more aggressively not work).

cc @ruchirK

### DIFF
--- a/src/operators/join.rs
+++ b/src/operators/join.rs
@@ -333,19 +333,14 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
 
             // Our initial invariants are that for each trace, physical compaction is less or equal the trace's upper bound.
             // These invariants ensure that we can reference observed batch frontiers from `_start_upper` onward, as long as
-            // we maintain our physical compaction capabilities appropriately.
+            // we maintain our physical compaction capabilities appropriately. These assertions are tested as we load up the
+            // initial work for the two traces, and before the operator is constructed.
 
-            // I can't say with confidence what should happen if this isn't the case, as it means we are unable to subset
-            // the current data without waiting for data that have not yet been created.
-            let mut trace1_start_upper = Antichain::new();
-            trace1.read_upper(&mut trace1_start_upper);
-            assert!(PartialOrder::less_equal(&trace1.get_physical_compaction(), &trace1_start_upper.borrow()));
-
-            let mut trace2_start_upper = Antichain::new();
-            trace2.read_upper(&mut trace2_start_upper);
-            assert!(PartialOrder::less_equal(&trace2.get_physical_compaction(), &trace2_start_upper.borrow()));
-
-            // acknowledged frontier for each input.
+            // Acknowledged frontier for each input.
+            // These two are used exclusively to track batch boundaries on which we may want/need to call `cursor_through`.
+            // They will drive our physical compaction of each trace, and we want to maintain at all times that each is beyond
+            // the physical compaction frontier of their corresponding trace.
+            // Should we ever *drop* a trace, these are 1. much harder to maintain correctly, but 2. no longer used.
             use timely::progress::frontier::Antichain;
             let mut acknowledged1 = Antichain::from_elem(<G::Timestamp>::minimum());
             let mut acknowledged2 = Antichain::from_elem(<G::Timestamp>::minimum());
@@ -363,6 +358,11 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
                 // `input1` with logic that would have otherwise been here. Check out the next loop
                 // for the structure.
             });
+            // At this point, `ack1` should exactly equal `trace1.read_upper()`, as they are both determined by
+            // iterating through batches and capturing the upper bound. This is a great moment to assert that
+            // `trace1`'s physical compaction frontier is before the frontier of completed times in `trace1`.
+            // TODO: in the case that this does not hold, instead "upgrade" the physical compaction frontier.
+            assert!(PartialOrder::less_equal(&trace1.get_physical_compaction(), &acknowledged1.borrow()));
 
             trace2.map_batches(|batch2| {
                 acknowledged2.clone_from(batch2.upper());
@@ -375,10 +375,15 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
                 // that property.
                 todo2.push_back(Deferred::new(trace1_cursor, trace1_storage, batch2_cursor, batch2.clone(), capability.clone()));
             });
+            // At this point, `ack2` should exactly equal `trace2.read_upper()`, as they are both determined by
+            // iterating through batches and capturing the upper bound. This is a great moment to assert that
+            // `trace2`'s physical compaction frontier is before the frontier of completed times in `trace2`.
+            // TODO: in the case that this does not hold, instead "upgrade" the physical compaction frontier.
+            assert!(PartialOrder::less_equal(&trace2.get_physical_compaction(), &acknowledged2.borrow()));
 
             // Droppable handles to shared trace data structures.
-            let mut trace1 = Some(trace1);
-            let mut trace2 = Some(trace2);
+            let mut trace1_option = Some(trace1);
+            let mut trace2_option = Some(trace2);
 
             // Swappable buffers for input extraction.
             let mut input1_buffer = Vec::new();
@@ -386,6 +391,8 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
 
             move |input1, input2, output| {
 
+                // 1. Consuming input.
+                //
                 // The join computation repeatedly accepts batches of updates from each of its inputs.
                 //
                 // For each accepted batch, it prepares a work-item to join the batch against previously "accepted"
@@ -399,7 +406,8 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
 
                 // Drain input 1, prepare work.
                 input1.for_each(|capability, data| {
-                    if let Some(ref mut trace2) = trace2 {
+                    // This test *should* always pass, as we only drop a trace in response to the other input emptying.
+                    if let Some(ref mut trace2) = trace2_option {
                         let capability = capability.retain();
                         data.swap(&mut input1_buffer);
                         for batch1 in input1_buffer.drain(..) {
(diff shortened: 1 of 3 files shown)


frankmcsherry:
@ryzhyk if you happen to have a test suite hanging around to point at this branch I'd be interested! We'll hammer on it at Materialize too, but you do different things than we do (e.g. multi-dimensional timestamps in your joins). The only intended change is that on pre-compacted arrangements it should be less wrong, and it may be that this hasn't been an issue for you, but we'll probably aim to land this or something like it.

ryzhyk:
> @ryzhyk if you happen to have a test suite hanging around to point at this branch I'd be interested! We'll hammer on it at Materialize too, but you do different things than we do (e.g. multi-dimensional timestamps in your joins). The only intended change is that on pre-compacted arrangements it should be less wrong, and it may be that this hasn't been an issue for you, but we'll probably aim to land this or something like it.

Sure, I'll get on it.  Is it possible to characterize the scenarios where the bug shows up and how it manifests outside of DD?

frankmcsherry:
Yeah, the main thing is that if you import an `Arranged` into a dataflow, where the underlying trace has undergone physical compaction, then .. under circumstances that now seem easy to reproduce (in Materialize) the compacted representation of the trace confuses `join` who produces double the output it should have (it takes in inputs A and B, and produces A x B in response to A and A x B in response to B).

It's a bit tricky to isolate, as it seems to reliably happen in the *second* execution of a Materialize query, and I haven't exactly figured out how to make it happen, but I can detail what does happen once it gets in that state.

frankmcsherry:
In maybe more detail: `join` starts up with the compacted trace, and in the first execution finds no inputs, but nonetheless learns that `[0, x)` is empty. It "accepts" this range and advances its internal bounds, and indicates to the trace that it will only need to be able to get access to the trace from `x` onward. However, the trace is perhaps compacted past `x` at this point. That should be an error there (and in this PR is now a `debug_assert!`) as you should not be able to roll back the physical compaction frontier. Instead, it overwrote the compaction frontier (it succeeded in rolling it back) and the next time around the loop asked for data up to that point, and got all the data instead (as the confused trace had more data, and yet saw that the request lined up with its new physical compaction frontier).

It's hard to explain without wincing at the sketchiness of it all. The new code is meant to be a lot more clear about stating, checking, and maintaining invariants.

ryzhyk:
Our tests are passing, but then they did not catch the bug in the first place...

frankmcsherry:
This has passed some amount of randomized stress testing at Materialize, so once we get a review on it I think we can land it (not erroring is good, but it should also land in a form that is more easy to understand than previously).

src/operators/join.rs
```diff
@@ -320,100 +320,127 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
...
+                // No `todo1` work here, because we haven't accepted anything into `batches2` yet.
+                // It is effectively "empty", because we choose to drain `trace1` before `trace2`.
+                // Once we start streaming batches in, we will need to respond to new batches from
+                // `input1` with logic that would have otherwise been here. Check out the next loop
+                // for the structure.
+            });
+
+            trace2.map_batches(|batch2| {
+                acknowledged2.clone_from(batch2.upper());
+                // It is safe to ask for `ack1` because we have confirmed it to be in advance of `distinguish_since`.
```
ruchirK, line 369: It's not clear to me where we confirmed that

Edit: it seems like `acknowledged1` will be the upper bound of the "most recent" batch in `trace1` and maybe that is how we know its in advance of `distinguish_since`?

src/operators/join.rs
```diff
@@ -320,100 +320,127 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
...
-                            if acknowledged2.is_none() { acknowledged2 = Some(timely::progress::frontier::Antichain::from_elem(<G::Timestamp>::minimum())); }
-                            if let Some(acknowledged2) = &mut acknowledged2 {
-                                if !PartialOrder::less_equal(&*acknowledged2, batch2.upper()) {
-                                    if !batch2.is_empty() {
-                                        panic!("Non-empty batch2 upper not beyond acknowledged frontier: {:?}, {:?}", batch2.upper(), acknowledged2);
-                                    }
+                                    todo2.push_back(Deferred::new(trace1_cursor, trace1_storage, batch2_cursor, batch2.clone(), capability.clone()));
                                 }
+
+                                // To update `acknowledged1` we might presume that `batch1.lower` should equal it, but we
```
ruchirK, line 441: I think this should be `acknowledged2` and `batch2`

src/operators/join.rs
```diff
@@ -320,100 +320,127 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
...
-                            if let Some(acknowledged1) = &mut acknowledged1 {
-                                if !PartialOrder::less_equal(&*acknowledged1, batch1.upper()) {
-                                        if !batch1.is_empty() {
-                                        panic!("Non-empty batch1 upper not beyond acknowledged frontier: {:?}, {:?}", batch1.upper(), acknowledged1);
-                                    }
-                                }
+                                // To update `acknowledged1` we might presume that `batch1.lower` should equal it, but we
+                                // may have skipped over empty batches. Still, the batches are in-order, and we should be
+                                // able to just assume the most recent `batch1.upper`
                                 acknowledged1.clone_from(batch1.upper());
```
ruchirK, line 419: should we maybe assert that acknowledged1 <= batch1.upper()? I feel like everything hinges on that (and we assert in `set_physical_compaction` so maybe its fine)

src/operators/join.rs
```diff
@@ -485,16 +514,16 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
 /// The structure wraps cursors which allow us to play out join computation at whatever rate we like.
 /// This allows us to avoid producing and buffering massive amounts of data, without giving the timely
 /// dataflow system a chance to run operators that can consume and aggregate the data.
-struct Deferred<K, V1, V2, T, R1, R2, R3, C1, C2, M, D>
```
ruchirK, line 488: the change here to not keep `mult` as a field seems like a unrelated refactor - is that right? or is there something more important going on?

src/operators/join.rs
```diff
@@ -451,28 +488,20 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
                 if trace2.is_some() && input1.frontier().is_empty() { trace2 = None; }
```
ruchirK, line 488: up to this point we did everything for {input,trace}1, then {input,trace}2 - this seems like the first instance where we do something for trace2 before trace1 - is the ordering important here?

src/operators/join.rs
```diff
@@ -451,28 +488,20 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
...
                     trace1.set_logical_compaction(input2.frontier().frontier());
-                    // At this point, if we haven't seen any input batches we should establish a frontier anyhow.
-                    if acknowledged1.is_none() {
-                        acknowledged1 = Some(Antichain::from_elem(<G::Timestamp>::minimum()));
-                    }
-                    if let Some(acknowledged1) = &mut acknowledged1 {
-                        trace1.advance_upper(acknowledged1);
-                        trace1.set_physical_compaction(acknowledged1.borrow());
-                    }
+                    trace1.advance_upper(&mut acknowledged1);
```
ruchirK, line 501: I don't quite understand how `set_logical_compaction`, `advance_upper` and `set_physical_compaction` interact (specifically the first two seem very overlapping?). Haven't read the docs on these so this is a bit of a lazy question - this could be a nice place for a few more comments

ruchirK (approved):
This seems like it makes sense! I don't fully understand everything this code is doing and I didn't take a close look to try to establish that the old / new versions of the code do the exact same stuff but I think the invariants we are trying to establish make sense. I left a few comments around places that seemed confusing.

src/operators/join.rs
frankmcsherry: Good point. We should be able to assert just after the loading, that the trace upper bounds are exactly the acknowledged frontiers. They are read out the same way (by scanning the existing batches), but that's not clear in this code.

src/operators/join.rs
frankmcsherry: Yes, oops.

src/operators/join.rs
frankmcsherry: We could! There would be a tremendous error if we minted batches that were not in-order, increasing. I'll put it in as a `debug_assert`; it's one of these things that has some run-time cost and should be `assert`ed elsewhere in the code.

src/operators/join.rs
frankmcsherry: It was related, in that .. now that we do that pre-work before starting the operator, we would try and stash `Deferred` in a vector in two places, and their closures were not the same. They were syntactically identical, but Rust makes them two separate types. So, we couldn't bake the multiplication into the type if we wanted to populate `todo2` in two place.

Fall-out from the main change, rather than a core change itself.

src/operators/join.rs
frankmcsherry: I think the ordering is not important, and it was original conceived as "handle `input1` being complete" but the logic hoists references to `trace2` up around it. I'll comment it up explaining the the order doesn't matter.

src/operators/join.rs
frankmcsherry: I'll add some more comments!

frankmcsherry:
I've restructured the code a bit, with substantially more comments, which I think resolve all the issues except perhaps the last comment about "how all these things interact". The comments are meant to address that, but I'd be up for any feedback on whether they do or do not do this.

src/operators/join.rs
```diff
@@ -484,24 +507,43 @@ impl<G, T1> JoinCore<G, T1::Key, T1::Val, T1::R> for Arranged<G,T1>
...
+                    if input2.frontier().is_empty() { trace1_option = None; }
+                    else {
+                        // Allow `trace1` to logically compact up to the frontier we may yet receive,
+                        // in the opposing input (`input2`). All `input2` times will be beyond this
+                        // frontier, and joined times only need to be accurate when advanced to it.
+                        trace1.set_logical_compaction(input2.frontier().frontier());
+                        // Allow `trace1` to physically compact up to the upper bound of batches we
+                        // have received in its input (`input1`). We will not require a cursor that
+                        // is not beyond this bound.
+                        trace1.set_physical_compaction(acknowledged1.borrow());
```
ruchirK, line 530: does acknowledged1 need to be <= input2.frontier() / is it always because of some other invariant?

frankmcsherry: No, they are generally unrelated.

PR #308, opened 2021-02-20
Clarify `Trace` capability nomenclature

frankmcsherry:
This PR changes `TraceReader` methods `advance_by` and `distinguish_since`, and their read counterparts, to more clearly named methods:
```rust
    fn set_logical_compaction(&mut self, frontier: AntichainRef<Self::Time>);
    fn get_logical_compaction(&mut self) -> AntichainRef<Self::Time>;
    fn set_physical_compaction(&mut self, frontier: AntichainRef<Self::Time>);
    fn get_physical_compaction(&mut self) -> AntichainRef<Self::Time>;
````
The documentation means to be a bit more clear about what these abilities entail.


frankmcsherry:
The prior method names are still present in `#[deprecated]` form, so ideally this doesn't break anything for anyone!

PR #319, opened 2021-03-23
Use bespoke traits in place of std::ops traits

frankmcsherry:
Previously, `difference.rs` built its group-theoretic traits out of `std::ops` traits like `AddAssign`, `Neg`, and `Mul`. These traits are not implemented for many common types that you might want, like tuples, arrays, and vectors. To use these types, custom structs were implemented, but with "miserable ergonomics".

This PR replaces the reliance on `std::ops` traits by moving the methods in to local traits with similar names, and expanded implementations.


benesch (approved):
LGTM. If you merge this I'll write the `derive` stuff, if you like!

PR #323, opened 2021-04-06
Add `join_function` operator

frankmcsherry:
This PR introduces the `join_function(logic)` linear operator on collections. The `logic` argument is a function from input data to *collections* i.e. a sequence of update triples. The result of `data.join_function(logic)` is effectively an equijoin between the contents of `data` and the contents of `(x, logic(x))` for all `x`. Rather than materialize that second collection, we just apply `logic` to each element of `x` and perform the appropriate corrections, namely joining timestamps and multiplying differences.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -270,6 +270,46 @@ impl<G: Scope, D: Data, R: Semigroup> Collection<G, D, R> where G::Timestamp: Da
             .as_collection()
     }
 
+    /// Joins each record against a collection defined by the function `logic`.
+    ///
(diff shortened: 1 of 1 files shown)


frankmcsherry:
NB this generalizes `map`, `filter`, `flat_map`, and `explode`. I don't know if it is a zero-cost abstraction, though I suppose it could be with `T::Minimum` and a difference of `1`. It is not a zero-cost cognitive burden, to think of those operators as joins against collection defined by functions, for sure.

PR #328, opened 2021-07-01
use `PartialOrder` implementation where appropriate

petrosagg:
These two places had a copy pasted implementation of the `PartialOrder` logic so this replaces it with calling the appropriate trait implementations

### DIFF
--- a/src/operators/arrange/arrangement.rs
+++ b/src/operators/arrange/arrangement.rs
@@ -569,7 +569,7 @@ where
                 *reader = Some(reader_local);
 
                 // Initialize to the minimal input frontier.
-                let mut input_frontier = vec![<G::Timestamp as Timestamp>::minimum()];
+                let mut prev_frontier = Antichain::from_elem(<G::Timestamp as Timestamp>::minimum());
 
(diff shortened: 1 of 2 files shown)


src/operators/arrange/arrangement.rs
```diff
@@ -589,14 +589,12 @@ where
...
-                    // to the new frontier).
-                    let progress = input_frontier.iter().any(|t2| !input.frontier().less_equal(t2));
-
-                    if progress {
+                    assert!(PartialOrder::less_equal(&prev_frontier.borrow(), &input.frontier().frontier()));
 
+                    // Test to see if strict progress has occurred, which happens whenever the new
+                    // frontier isn't equal to the previous. It is only in this case that we have any
+                    // data processing to do.
+                    if prev_frontier.borrow() != input.frontier().frontier() {
```
frankmcsherry, line 597: I think this is right. I was a bit confused about the definition of antichain equality, but it looks like the implementation equates two things that have the same elements but in different orders.

frankmcsherry (approved):
This looks good to me, thanks!

PR #332, opened 2021-07-27
Extract arrangement's exchange fn, pass pact as parameter

antiguru:
Extend Differential's API to take a PACT in more places to enable
clients picking their own implementation. Export the default exchange
logic as a function.


antiguru:
I don't think this is needed anymore. I .. can't recall why this was useful :D

PR #334, opened 2021-08-04
Avoid resetting cursor on no-op seek

frankmcsherry:
The implementation of `seek` had the defective property that when the seek was a no-op, for example when one re-issued a seek to the same key multiple times in sequence, the child cursor (over associated values, say) would be reset each time. I judge the correct implementation of seek to not do this, so as to present a continually advancing cursor. One can always rewind a cursor's values if that is what you want. (of course, one could always test if you are at the correct key and not `seek` as well).

### DIFF
--- a/src/trace/layers/ordered.rs
+++ b/src/trace/layers/ordered.rs
@@ -262,8 +262,11 @@ where
         }
     }
     fn seek(&mut self, storage: &OrderedLayer<K, L, O>, key: &Self::Key) {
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Counterpoint: it is perhaps antisocial for a `seek` call not to result in a specifically positioned cursor, but rather one that may need to be checked to see if it has already been slightly advanced (logic that wants to enumerate values for keys with repeats may need to be smarter now, or risk having errors).

PR #335, opened 2021-08-05
trace: add PartialEq/Eq impls for Description

ruchirK:


frankmcsherry (approved):
Looks great, thanks very much!

PR #336, opened 2021-08-11
merge_batcher: Base buffer size off 8KiB

antiguru:
This PR changes the allocation policy in `merge_batcher` to not use 1024-element vectors but rather fit the vectors in 8KiB buffers.

@frankmcsherry Do the buffers need to have a power-of-two size? If yes, the logic to calculate the buffer length needs to be changed.


frankmcsherry:
This also puts empty vectors out there, and I think will result in repeated doubling of allocations as things are filled. Is that intended as well? (the 8KB thing seems good, the empty vectors seem more contentious).

frankmcsherry:
Looks good to me (and cleans up the other random popping calls with abstraction; excellent!)

PR #339, opened 2021-08-24
trace: let set_{logical,physical}_compaction take arguments equal to …

ruchirK:
…or beyond current frontier.

This commit changes the requirements  set_{logical,physical}_compaction methods to match
those of Capability::downgrade. Now, these methods all accept arguments equal to or beyond
the most recent argument passed, or initial value if the function was never called before.

Also added a debug assert to make sure the logical compaction only advances.


ruchirK:
cc: @frankmcsherry this is the thing we discussed on Monday. I looked through the code for callers of `set_logical_compaction` / `set_physical_compaction` -- I didn't find anything that enforced the old strictly greater than invariant.

PR #340, opened 2021-08-31
Relax set_*_compaction methods

frankmcsherry:
The various `TraceAgent::set_*_compaction` methods would assert if certain invariants were not satisfied. This clashed a bit with their existing uses which didn't incorporate the initial state of the `TraceAgent` into their reasoning, resulting in panics. Uses of these methods with frontiers that are not greater-equal the current compaction frontier are "in error", but this behavior may be more generous and less crashy. Any user can assert that their argument to this call equals the result of `TraceAgent::get_*_compaction` and crash on their own if they prefer, but DD doesn't need to force them to crash.


frankmcsherry:
Fixes https://github.com/MaterializeInc/materialize/issues/8119

PR #341, opened 2021-09-10
Memory alloc improvements

antiguru:
A collection of small improvements to bypass the memory allocator when possible.

### DIFF
--- a/src/lattice.rs
+++ b/src/lattice.rs
@@ -222,12 +222,45 @@ implement_lattice!((), ());
 /// ```
 pub fn antichain_join<T: Lattice>(one: &[T], other: &[T]) -> Antichain<T> {
     let mut upper = Antichain::new();
+    antichain_join_into(one, other, &mut upper);
+    upper
+}
+
+/// Returns the "smallest" minimal antichain "greater or equal" to both inputs.
+///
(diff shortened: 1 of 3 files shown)


src/trace/mod.rs
```diff
@@ -166,7 +166,8 @@ pub trait TraceReader {
         target.clear();
         target.insert(<Self::Time as timely::progress::Timestamp>::minimum());
         self.map_batches(|batch| {
-            target.clone_from(batch.upper());
+            target.clear();
+            target.extend(batch.upper().iter().cloned());
```
frankmcsherry, line 170: This is a bit surprising as an improvement. I think `clone_from()` is supposed to do this already.

antiguru: I removed this change as it should be superseded by https://github.com/frankmcsherry/timely-dataflow/commit/7f5b8b75651d48380c3183195fb661383f660096

frankmcsherry: For posterity: `clone_from()` does not do the expected thing; the derived implementation calls `clone()` rather than `clone_from()`.

frankmcsherry (approved):
These look great, thanks!

PR #342, opened 2021-10-05
Allow halfjoin to yield tastefully

frankmcsherry:
This PR modifies some internals for `halfjoin` to allow it to yield by user-supplied logic. That logic can look at the elapsed time and number of matched records, where ideally if it uses neither of them there is not much overhead (the time is provided as an `Instant` which means that the user decides to call `elapsed` or not).

cc @uce

### DIFF
--- a/dogsdogsdogs/src/operators/half_join.rs
+++ b/dogsdogsdogs/src/operators/half_join.rs
@@ -95,7 +95,7 @@ where
         let dout = (output_func(k, v1, v2), time.clone());
         Some((dout, initial.clone(), diff))
     };
-    half_join_internal_unsafe(stream, arrangement, frontier_func, comparison, output_func)
+    half_join_internal_unsafe(stream, arrangement, frontier_func, comparison, |timer, count| false, output_func)
 }
 
 /// An unsafe variant of `half_join` where the `output_func` closure takes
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Thanks! This looks fine

dogsdogsdogs/src/operators/half_join.rs
```diff
@@ -154,81 +160,106 @@ where
...
+                                cursor.seek_key(&storage, &key);
+                                if cursor.get_key(&storage) == Some(&key) {
+                                    while let Some(val2) = cursor.get_val(&storage) {
+                                        cursor.map_times(&storage, |t, d| {
+                                            if comparison(t, initial) {
+                                                output_buffer.push((t.join(time), d.clone()))
+                                            }
+                                        });
+                                        consolidate(&mut output_buffer);
+                                        work += output_buffer.len();
```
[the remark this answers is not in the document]
frankmcsherry: Good point. Added more to the method documentation.

dogsdogsdogs/src/operators/half_join.rs
```diff
@@ -154,81 +161,105 @@ where
...
-                    let mut session = output.session(capability);
+            if let Some(ref mut trace) = arrangement_trace {
 
-                    // Sort requests by key for in-order cursor traversal.
-                    consolidate_updates(proposals);
+                for (capability, proposals) in stash.iter_mut() {
 
-                    let (mut cursor, storage) = trace.cursor();
+                    // Avoid computation if we should already yield.
+                    // TODO: Verify this is correct for TOTAL ORDER.
```
uce, line 198: @frankmcsherry Did you verify this?

frankmcsherry: Not yet. This is an old comment though. I think it is green because the indentation level changed.

PR #352, opened 2022-02-18
use "in advance of" terminology instead of "beyond"

petrosagg:


petrosagg:
@Kixiron not sure why the downvote but the "beyond" term is something that is used in materialize instead of "in advance of" for.. reasons and has crept in DD because the same people work in both. If you grep the repo for "in advance of" you'll see that it is already in wide use.

Kixiron:
I think beyond is more succinct and feels less artificial really, plus they mean the same thing so it's not a loss of information

PR #353, opened 2022-02-20
consolidation: fix undefined behaviour from mutable pointer aliasing

petrosagg:
Running the added testcase under miri without the equality check results in Undefined Behaviour from creating two mutable references pointing to the same location.

### DIFF
--- a/src/consolidation.rs
+++ b/src/consolidation.rs
@@ -66,7 +66,7 @@ pub fn consolidate_slice<T: Ord, R: Semigroup>(slice: &mut [(T, R)]) -> usize {
                     offset += 1;
                 }
                 let ptr1 = slice.as_mut_ptr().offset(offset as isize);
-                std::mem::swap(&mut *ptr1, &mut *ptr2);
+                std::ptr::swap(ptr1, ptr2);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Can you talk me through why it isn't a false positive? The loop invariant is that `offset < index`, and if that isn't true there is a bigger problem. Does the test case demonstrate a violation of this invariant?
```rust
            // LOOP INVARIANT: offset < index
            let ptr1 = slice.as_mut_ptr().offset(offset as isize);
            let ptr2 = slice.as_mut_ptr().offset(index as isize);
```

petrosagg:
Yeah the invariant is violated in the else statement with the test case I added

frankmcsherry:
So I think I figured out what I meant by "in the wrong order" above. The consolidation logic switches from "validating" in which all adjacent elements are distinct, to "compacting" as soon as two elements have equal keys. In the former no movement needs to happen, and in the latter the swapping is always between distinct locations.

If you don't mind, I'll write that up, and then maybe we can check that it passes the test?

petrosagg:
@frankmcsherry I updated the code to a version that doesn't have an extra check

petrosagg:
ok I put up a final version that replaces `std::mem::swap` with `std::ptr::swap` which is ok with aliasing mut pointers

frankmcsherry (approved):
Looks great, thanks very much!

PR #354, opened 2022-02-20
Consolidation now skips distinct, non-zero updates

frankmcsherry:
Changes consolidation logic to have it scan over distinct, non-zero updates to start, essentially seeking the first moment where consolidation may occur. At that point, the logic becomes more complicated, but if we can skip it until then, great!

Also fixes a bug noticed by @petrosagg, but more obliquely than his recommendation (use `std::ptr::swap`).

### DIFF
--- a/src/consolidation.rs
+++ b/src/consolidation.rs
@@ -40,21 +40,27 @@ pub fn consolidate_slice<T: Ord, R: Semigroup>(slice: &mut [(T, R)]) -> usize {
 
     // Counts the number of distinct known-non-zero accumulations. Indexes the write location.
     let mut offset = 0;
-    for index in 1 .. slice.len() {
 
-        // The following unsafe block elides various bounds checks, using the reasoning that `offset`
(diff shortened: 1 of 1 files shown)


src/consolidation.rs
```diff
@@ -40,21 +40,27 @@ pub fn consolidate_slice<T: Ord, R: Semigroup>(slice: &mut [(T, R)]) -> usize {
 
     // Counts the number of distinct known-non-zero accumulations. Indexes the write location.
     let mut offset = 0;
-    for index in 1 .. slice.len() {
 
-        // The following unsafe block elides various bounds checks, using the reasoning that `offset`
+    // We can first move quickly through any prefix of distinct, non-zero elements
+    while offset + 1 < slice.len() && !(slice[offset].0 == slice[offset+1].0) && !slice[offset].1.is_zero() {
```
petrosagg, line 45: shouldn't this happen inside the loop so that we skip over all the distinct regions rather than just the first one?

PR #355, opened 2022-02-20
count: add generic `count_total_core` method

petrosagg:
This follows the pattern of `distinct_core` to allow counting a collection but use a different diff type than `isize`.


frankmcsherry:
Looks good, tyvm!

PR #360, opened 2022-03-17
trace: define `BatchReader` with associated types

petrosagg:
The `TraceReader` trait uses associated types to define its `Key`, `Val`, `Time`, `Diff` but the `BatchReader` trait did not, even though they are very similar in nature. Usually the choice between associated types or generic parameters on a trait is determined by whether or not a particular type is expected to implement the same trait multiple times.

My starting point was that these two trait should at the very least be consistent with respect to their structure and either both use generic parameters or both use associated types.

All the uses in this repo (and also that I can imagine being useful) don't really need `BatchReader` to be polymorphic for a particular type and so I chose to change that one to make it consistent with `TraceReader`.

The result is quite pleasing as in many cases a lot of generic parameters are erased.

In order to keep this PR short I left the `Cursor` trait untouched, but I believe a similar transformation would be beneficial there too, simplifying further many type signatures.


petrosagg:
superceded by #367

PR #361, opened 2022-03-18
remove transmute and fix rust warning

petrosagg:
while `()` is a ZST the potentially dangling reference is still undefined behaviour. Making it a static is a trivial fix.


petrosagg:
superceded by #367

PR #363, opened 2022-03-30
trace: remove redundant bounds implied by the trait definition

petrosagg:
Many methods in DD are requiring the `BatchReader` and `Cursor` bounds on the `Batch` and `Cursor` associated types of a `TraceReader` respectively. However these bounds are already implied by the trait definition and can be omitted, resulting in simpler signatures.


frankmcsherry:
Did Rust land some `implied_bounds` work? These bounds have traditionally been there because Rust complains if they are not. If the world changed while I wasn't looking, hooray!

petrosagg:
Hm, I'm not sure. I haven't tried compiling with an older rust

petrosagg:
superseded by #367

PR #364, opened 2022-04-25
make capture::source's token thread-safe

guswynn:
As part of resolving: https://github.com/MaterializeInc/materialize/pull/11987, we 

in general, thread-safety comes at a small perf cost, but in this case, the sync activator is only called once, and the Arc is never cloned, so it will not be a measurable difference, other than the `upgrade` costing one atomic load each run. This is a strictly more general api now! The `Send + Sync` can be coerced away if need be

I can clean this up once I make a Sync version of `ActivatorOnDrop` in the timely repo!

tested locally, token drop does in fact clear the local token vector

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -359,66 +359,65 @@ pub mod source {
         let token_mut = &mut token;
         messages_op.build(move |capabilities| {
 
-            // Vector of strong references to capabilities, which can be dropped to terminate the sources.
+            // A Weak that communicates whether the returned token has been dropped.
             let drop_activator_weak = Arc::downgrade(&drop_activator);
 
             *token_mut = Some(drop_activator);
 
             // Read messages from some source; shuffle them to UPDATES and PROGRESS; share capability with FEEDBACK.
-            // First, wrap capabilities in a rc refcell so that they can be downgraded to weak references.
-            let capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
-            let capability_sets = ActivateOnDrop::new(capability_sets, Rc::new(address), activations);
-            let mut local_capabilities = Some(capability_sets);
+            let mut capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
             // Capture the shared frontier to read out frontier updates to apply.
             let local_frontier = shared_frontier.clone();
             //
             move |_frontiers| {
+                // First check to ensure that we haven't been terminated by someone dropping our tokens.
                 if drop_activator_weak.upgrade().is_none() {
-                    local_capabilities = None;
+                    // Give up our capabilities
+                    capability_sets.0.downgrade(&[]);
+                    capability_sets.1.downgrade(&[]);
+                    // never continue, even if we are (erroneously) activated again.
+                    return;
                 }
 
-                // First check to ensure that we haven't been terminated by someone dropping our tokens.
-                if let Some(capabilities) = &mut local_capabilities {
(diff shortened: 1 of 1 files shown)


src/capture.rs
```diff
@@ -338,23 +352,36 @@ pub mod source {
...
+
             // Read messages from some source; shuffle them to UPDATES and PROGRESS; share capability with FEEDBACK.
             // First, wrap capabilities in a rc refcell so that they can be downgraded to weak references.
             use timely::scheduling::activate::ActivateOnDrop;
             let capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
             let capability_sets = ActivateOnDrop::new(capability_sets, Rc::new(address), activations);
             let strong_capabilities = Rc::new(RefCell::new(capability_sets));
             let local_capabilities = Rc::downgrade(&strong_capabilities);
-            tokens_mut.push(strong_capabilities);
+            tokens.push(strong_capabilities);
```
frankmcsherry, line 375: If we don't intend to return `tokens`, I think we should remove it as a concept and just hold on to `capability_sets`.

guswynn: updated!

src/capture.rs
```diff
@@ -338,26 +352,34 @@ pub mod source {
...
+            // Vector of strong references to capabilities, which can be dropped to terminate the sources.
+            let drop_activator_weak = Arc::downgrade(&drop_activator);
+
+            *token_mut = Some(drop_activator);
+
             // Read messages from some source; shuffle them to UPDATES and PROGRESS; share capability with FEEDBACK.
             // First, wrap capabilities in a rc refcell so that they can be downgraded to weak references.
-            use timely::scheduling::activate::ActivateOnDrop;
             let capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
             let capability_sets = ActivateOnDrop::new(capability_sets, Rc::new(address), activations);
```
frankmcsherry, line 370: Should this depart as well?

src/capture.rs
```diff
@@ -338,26 +352,34 @@ pub mod source {
...
-            let strong_capabilities = Rc::new(RefCell::new(capability_sets));
-            let local_capabilities = Rc::downgrade(&strong_capabilities);
-            tokens_mut.push(strong_capabilities);
+            let mut local_capabilities = Some(capability_sets);
```
frankmcsherry, line 371: I think you can just leave them as sets, and call `downgrade([])` rather than setting the option to `None`.

src/capture.rs
```diff
@@ -359,66 +359,65 @@ pub mod source {
...
             let drop_activator_weak = Arc::downgrade(&drop_activator);
 
             *token_mut = Some(drop_activator);
 
             // Read messages from some source; shuffle them to UPDATES and PROGRESS; share capability with FEEDBACK.
-            // First, wrap capabilities in a rc refcell so that they can be downgraded to weak references.
-            let capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
-            let capability_sets = ActivateOnDrop::new(capability_sets, Rc::new(address), activations);
-            let mut local_capabilities = Some(capability_sets);
+            let mut capability_sets = (CapabilitySet::from_elem(capabilities[0].clone()), CapabilitySet::from_elem(capabilities[1].clone()));
```
frankmcsherry, line 368: Nit, but can I trick you into breaking this out as `update_caps` and `progress_caps`?

frankmcsherry (approved):
Looks great, thanks!

PR #367, opened 2022-05-23
Further tidying up of submitted PRs

petrosagg:
@frankmcsherry thanks for reviewing #360 #361 and #363! I saw you incorporated it them all into #366 but the commits lost provenance/got mixed with other changes of yours.

I opened this PR that has a zero-diff with #366 but with the original commits preserved and the additional changes added by you on top. I looked at the additional changes and all looked good to me! The only question I had was regarding the manual `Clone` implementations, which I've split out in the last commit of this PR. They look identical to what the automatically derived implementation would do, is it a stylistic preference or am I missing something?

I'll close #360 #361 and #363 in hopes of getting this one merged :crossed_fingers:


frankmcsherry:
If things look confusing, it is because I started from just your PR descriptions rather than the PRs themselves. I knew I wanted to hit a few more things than you did (e.g. simplifying generic arguments to just the `Batch` argument, and doing the same thing for `Cursor`), so I retyped all of it. I'm happy to reframe this as edits on your work though, now that this experiment has been conducted (and I understand more clearly why various changes are made / not made).

The manual `Clone` implementations are .. I think just a consequence of me not doing that optimization once I removed the generic arguments (and .. perhaps there are others now too?). Any other weird edits aren't opinions on my part so much as diffs between what you and I did (I did minimal reformatting, so probably those edits are just returning toward the original state; could take or leave them!).

frankmcsherry:
Thanks much, and apologies for the confusion!

PR #382, opened 2023-03-12
Inline functions frequently called from Materialize

antiguru:
Observed by applying callgrind on Materialize.


antiguru (commented):
@frankmcsherry LGTM'ed this, thanks!

antiguru (commented):
@frankmcsherry LGTM'ed this, thanks!

PR #383, opened 2023-03-17
Apply changes from Timely TimelyDataflow/timely-dataflow#515

antiguru:
DO NOT MERGE

TimelyDataflow/timely-dataflow#515 breaks some external APIs and this PR shows what needs to change in Differential.


antiguru:
Not needed currently, closing.

PR #387, opened 2023-04-14
Expose TraceAgent::operator

antiguru:
Expose access to `TraceAgent::operator` to provide access to the underlying `OperatorInfo`.


frankmcsherry (approved):
Thanks for the contribution!

PR #388, opened 2023-04-15
Fix `find_cycles` in the progress example

nooberfsh:


frankmcsherry (approved):
This seems like a great fix; sorry about the bug!

PR #391, opened 2023-05-26
Failed attempt at supporting TimelyStack for (T, R) in leafs

antiguru:
DO NOT MERGE

This shows where we cannot use `TimelyStack<(T, R)>` to store time-diff pairs in leafs because `advance_builder_from` requires mutable access to the contents.


antiguru:
Moritz to file an issue about this.

PR #392, opened 2023-06-05
join: simplify logic passed to `Deferred::work`

teskje:
Previously, the join code had to wrap the `result` closure into another closure to invoke `Deferred::work`, to deal with the fact that matches from the `todo1` list had their value and diff fields swapped.

Instead of having a closure to unswap the fields, we can simply pass them to the `Deferred` constructor in the correct order instead. That is, the `Deferred` constructor should always receive the cursor/storage from input 1 first and from input 2 second. If we make this change, then a) no swapping closure is necessary, and b) the input values match the types (`C1`, `C2`) of the `Deferred` struct.

This PR is a backport of https://github.com/MaterializeInc/materialize/pull/19667, to avoid that the implementations diverge unnecessarily. In `mz_join_core`, the readability improvement from this change is more pronounced than in DD.


frankmcsherry:
This seems like a good idea. I'm not sure why the code is the other way!

frankmcsherry:
This has been corrected in the current codebase, with the join tactics embodying the unification presented here. Thanks for raising the issue though!

PR #399, opened 2023-06-28
Spine drop batches once logical and physical compaction frontiers are empty

### LINKED ISSUE #398: Holding on to a trace with physical/logical compaction to the empty frontier stalls compaction
(Placeholder, @antiguru to add more info+repro)

While working on https://github.com/MaterializeInc/materialize/pull/18766, we noticed that allowing physical and logical compaction of a trace to the empty frontier prevents compaction, even if other owners of the trace eventually allow compaction. It is required to repeatedly allow compaction with a sensible frontier (i.e., `read_upper` followed by allowing compaction) to enable compaction of the spine.

antiguru:
Once the ~physical~ logical compaction frontier is empty, there are no times that can be represented in a a trace, which means that we're good to drop all contents.

This fixes #398.

### DIFF
--- a/src/trace/implementations/spine_fueled.rs
+++ b/src/trace/implementations/spine_fueled.rs
@@ -212,18 +212,9 @@ where
     }
     #[inline]
     fn set_logical_compaction(&mut self, frontier: AntichainRef<B::Time>) {
-        // If the logical frontier is empty, we are not obliged to be able to represent any time,
-        // since no time is in advance of `[]`. In the case, we can drop all contents.
-        //
-        // We need to call `close` before advancing `logical_frontier` because otherwise we cannot
-        // insert anymore.
-        if frontier.is_empty() {
-            self.drop_batches();
-            self.close();
-        }
-
         self.logical_frontier.clear();
         self.logical_frontier.extend(frontier.iter().cloned());
+        self.consider_closing();
     }
     #[inline]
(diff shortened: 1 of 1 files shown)


frankmcsherry:
I don't think this is right. Physical compaction can be allowed up through `[]` and it just means that you are only allowed to ask for a cursor through `[]`, not that the trace is invalidated. Logical compaction would be the thing that when it reaches `[]` would mean that the trace is not obliged to be correct for any time (as none are in advance of / beyond `[]`).

src/trace/implementations/spine_fueled.rs
```diff
@@ -482,6 +482,12 @@ where
             }
         }
 
+        // If the logical frontier is empty, we are not obliged to be able to represent any time,
+        // since no time is in advance of `[]`. In the case, we can drop all contents.
+        if self.logical_frontier.is_empty() {
+            self.drop_batches();
+        }
+
```
frankmcsherry, line 490: Is there an explainer for why the code is here, rather than in e.g. `set_logical_compaction`?

antiguru: Moved the code to `set_logical_compaction` and `insert`. As far as I can tell these are the only places where logical compaction is advanced, and new batches are inserted.

antiguru:
Closing because this approach is not correct.

PR #401, opened 2023-07-04
Downgrade/upgrade trace agents

antiguru:
Add the ability to downgrade trace agents to only hold weak references to the trace. The weak reference can be upgraded and dereferenced to the trace, but not upgraded to a full agent. This is because the caller already gave up the right to prevent compaction, and re-acquiring it, while technically possible, introduces uncertainty about what time can be selected.

## Example

```rust
let weak_trace = arranged.trace.downgrade();

if let Some(trace) = weak_trace.upgrade() {
    trace.as_deref().map_batches(|_| {});
}
```

### DIFF
--- a/src/operators/arrange/agent.rs
+++ b/src/operators/arrange/agent.rs
@@ -160,6 +160,17 @@ where
     pub fn operator(&self) -> &OperatorInfo {
         &self.operator
     }
+
+    /// Downgrade the trace handle.
+    ///
+    /// The weak trace agent can be upgraded to inspect the trace if there are other agents still holding
+    /// on to the trace. The weak handle does not prevent logical and physical compaction, and cannot
+    /// be upgraded to a [`TraceAgent`].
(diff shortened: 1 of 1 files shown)


frankmcsherry (commented):
I think there are some design / naming choices to revisit here. I can see what is going on, but for example what you've created isn't a weak trace agent, it is a different thing that only has access to the underlying trace and not e.g. `queue`, risks exposing mutable access to `trace` (it doesn't in this PR, but someone needs to know that you should not expose a `RefMut` of `Tr` or risk violating frontier invariants should anyone call `set_X_compaction()`).

Let me take a ponder about how to better encapsulate "compaction-ensured trace access" and "subscription to trace" as two things. Personally, I prefer fixing the other bug around `[]` frontiers, and using that as the mechanism to provide weak traces rather than a parallel alternative.

antiguru:
We implemented an alternative in #404.

Issue #190, opened 2019-07-22
Export timely dataflow


frankmcsherry:
For various headache reasons, it may make sense for differential to export the version of timely dataflow on  which it depends. This has the potential to relief pains felt by version slew, where persons who care not at all which version of timely is required can simply use the re-exported timely from differential.

cc @benesch

PR #403, opened 2023-07-18
re-export timely

discord9:
#190 suggest so, and it did seem cause less headache


discord9:
Do you think it's wise to export timely here? @frankmcsherry

PR #404, opened 2023-08-01
Provide TraceAgent::trace_box_unstable to expose the inner trace

antiguru:
This method exposes the trace agent's trace, which requires the caller to maintain the invariants expected by Differential. Specifically, the caller should not mutate the structure, and should not hold on to strong references. The method is not part of Differential's (semi-) stable API.


frankmcsherry (approved):
Looks great, and sorry for the delays!

PR #405, opened 2023-08-18
Shrink containers if their capacity isn't needed

antiguru:
This eagerly releases memory after retracting large amount of data, without having to wait for another operation to happen.


antiguru:
With https://github.com/TimelyDataflow/timely-dataflow/pull/530, we could shrink the Columnation-based variant, too.

antiguru:
Not needed anymore with `ord_neu`. In addition, this only releases virtual memory, so it might need more memory than before because it needs to copy into a new allocation.

PR #407, opened 2023-08-28
Efficient merging with empty batches

teskje:
This PR makes merging of batches with empty batches efficient. Previously this operation would be handled like a regular merge, by allocating a new batch and merging the input batches into it. However, if one of the batches is empty, there is no need to allocate a new batch, we can simply extend the bounds of the existing non-empty batch accordingly.

The implementation of this optimization is complicated by the fact that there are `Batch` implementations for `Rc<Batch>` and `Abomonated<Batch>`. Both of these only provide an immutable reference to the underlying batch, making it impossible to adjust the batch description to extend the batch's bounds. This commit solves this by introducing new `RcBatch` and `AbomonatedBatch` wrappers that, in addition to the `Rc`/`Abomonated` batch, hold their own `Description`s that override the inner batch's description.

Some adjustments to logging of merge events is also necessary, to account for the new merge path.

Fixes https://github.com/MaterializeInc/materialize/issues/21165.


teskje:
[Materialize CI](https://github.com/MaterializeInc/materialize/pull/21407) is happy with this change.

PR #408, opened 2023-10-18
reduce: remove superfluous cloning of keys

petrosagg:
Just a small thing I noticed while reading the code


frankmcsherry (approved):
Hrm. Seems good!

PR #411, opened 2023-11-07
Allow custom exertion logic

frankmcsherry:
This PR allows users to set their own "exertion logic", replacing the hardwired `reduced` function for spines. This was the method that indicated whether there was work to perform in the absence of actual updates. This PR combines this with the `effort` argument to allow logic that produces an `Option<usize>` indicating how many updates should be virtually applied, as a function of the shape of the spine. It's a bit of a leaky abstraction (not all traces need to have a spine, right?) but the moment at which an opinion is currently injected requires the trace abstraction.

Some follow-on thoughts:
1. We could change the `Option<usize>` output to `usize`, and commit to `0` to mean "do no work, and do not reschedule".
2. We could change the `idle_merge_effort` config to be `usize` rather than `isize`, which I think it is because ultimately the `exert` function needs that. But it doesn't make a lot of sense to provide negative idle merge effort.
3. We could try and put the method on a `Spine` type, rather than as part of `Trace`, but I'm not certain how to thread the system-wide config of "do this with a spine" other than across the trace.

cc: @antiguru, @teskje

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,5 +140,17 @@ impl Config {
 pub fn configure(config: &mut timely::WorkerConfig, options: &Config) {
     if let Some(effort) = options.idle_merge_effort {
         config.set("differential/idle_merge_effort".to_string(), effort);
+        config.set::<trace::ExertionLogic>(
+            "differential/default_exert_logic".to_string(),
+            std::sync::Arc::new(move |batches| {
+                let mut non_empty = 0;
+                for (_index, count, length) in batches {
+                    if count > 1 { return Some(effort as usize); }
+                    if length > 0 { non_empty += 1; }
+                    if non_empty > 1 { return Some(effort as usize); }
+                }
+                None
+            }),
+        );
     }
 }
--- a/src/trace/implementations/spine_fueled.rs
+++ b/src/trace/implementations/spine_fueled.rs
@@ -489,8 +489,8 @@ where
             }
         }
 
-        // Having performed all of our work, if more than one batch remains reschedule ourself.
-        if !self.exert_effort().is_some() {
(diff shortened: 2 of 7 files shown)


src/trace/implementations/spine_fueled.rs
```diff
@@ -483,7 +490,7 @@ where
         }
 
         // Having performed all of our work, if more than one batch remains reschedule ourself.
-        if !self.reduced() {
+        if !self.exert_effort().is_some() {
```
antiguru, line 493:
```suggestion
        if self.exert_effort().is_none() {
```

src/lib.rs
```diff
@@ -140,5 +140,17 @@ impl Config {
 pub fn configure(config: &mut timely::WorkerConfig, options: &Config) {
     if let Some(effort) = options.idle_merge_effort {
         config.set("differential/idle_merge_effort".to_string(), effort);
```
antiguru, line 142: I don't think we're referencing `idle_merge_effort` anymore, so we should probably remove it here, and extend `options` to also take a `default_excert_logic`.

antiguru (approved):
Looks good! Left some comments inline.

src/trace/implementations/spine_fueled.rs
frankmcsherry: Ooh, I think this is wrong both ways, actually. We want to reschedule if there is exert effort to apply. Oops. Will fix.

src/lib.rs
frankmcsherry: I think I'm up for waiting just a bit to see how this looks before locking things in. I wrote a bunch of code before realizing that it all needs to cross thread boundaries, and if it turns out that it needs to be SerDe also .. what a mess. :D

PR #412, opened 2023-11-08
Get rid of some UB

umanwizard:
Constructing a reference to an invalid value is always UB, according to [this documentation](https://doc.rust-lang.org/nomicon/what-unsafe-does.html).

In addition, even just the `set_len` call can cause UB if we panic immediately after it (and thus try to drop the invalid value).

The new implementation is taken from `Vec::push`. Hat tip to @guswynn , @doy-materialize, and @petrosagg for talking me through why it was indeed UB, and the latter for also suggesting the `Vec::push` implementation.


frankmcsherry:
Discussed offline, but we merged instead #413 which simply removes the `unsafe` at the cost of some bounds checks. But I do appreciate the attention to sloppiness on my part, and ways to correct it!

PR #413, opened 2023-11-09
Remove push_unchecked

antiguru:
On recent Rust compilers, it produces the same assembly as `Vec::push`, so there is no benefit in keeping the function around, besides that it avoids occasionally nerd-sniping people in trying to understand Rust's UB and how to write the function correctly.


frankmcsherry (approved):
Down with `unsafe`!

PR #418, opened 2023-11-14
Another attempt at introducing a columnated merge batcher

antiguru:
This builds on #414 but uses the new Layout feature recently merged.

### DIFF
--- a/src/trace/implementations/merge_batcher_col.rs
+++ b/src/trace/implementations/merge_batcher_col.rs
@@ -120,13 +120,11 @@ struct TimelyStackQueue<T: Columnation> {
 
 impl<T: Columnation> Default for TimelyStackQueue<T> {
     fn default() -> Self {
-        Self::empty()
+        Self::from(Default::default())
     }
 }
 
 impl<T: Columnation> TimelyStackQueue<T> {
-
-    fn empty() -> Self { TimelyStackQueue::from(Default::default()) }
 
     fn pop(&mut self) -> &T {
         self.head += 1;
@@ -145,7 +143,6 @@ impl<T: Columnation> TimelyStackQueue<T> {
     }
 
     fn done(mut self) -> TimelyStack<T> {
-        self.list.clear();
         self.list
     }
 
@@ -285,8 +282,8 @@ impl<D: Ord+Clone+Columnation+'static, T: Ord+Clone+Columnation+'static, R: Semi
         let mut list1 = list1.into_iter();
         let mut list2 = list2.into_iter();
 
-        let mut head1 = list1.next().map(TimelyStackQueue::from).unwrap_or_default();
-        let mut head2 = list2.next().map(TimelyStackQueue::from).unwrap_or_default();
+        let mut head1 = TimelyStackQueue::from(list1.next().unwrap_or_default());
+        let mut head2 = TimelyStackQueue::from(list2.next().unwrap_or_default());
 
         // while we have valid data in each input, merge.
         while !head1.is_empty() && !head2.is_empty() {
@@ -320,11 +317,11 @@ impl<D: Ord+Clone+Columnation+'static, T: Ord+Clone+Columnation+'static, R: Semi
 
             if head1.is_empty() {
                 self.recycle(head1.done());
-                head1 = list1.next().map(TimelyStackQueue::from).unwrap_or_default();
+                head1 = TimelyStackQueue::from(list1.next().unwrap_or_default());
             }
             if head2.is_empty() {
                 self.recycle(head2.done());
-                head2 = list2.next().map(TimelyStackQueue::from).unwrap_or_default();
+                head2 = TimelyStackQueue::from(list2.next().unwrap_or_default());
             }
         }
 
--- a/src/trace/implementations/merge_batcher_col.rs
+++ b/src/trace/implementations/merge_batcher_col.rs
@@ -1,5 +1,6 @@
 //! A general purpose `Batcher` implementation based on radix sort for TimelyStack.
 
+use std::marker::PhantomData;
 use timely::Container;
 use timely::communication::message::RefOrMut;
 use timely::container::columnation::{Columnation, TimelyStack};
@@ -21,7 +22,7 @@ pub struct ColumnatedMergeBatcher<B: Batch>
     sorter: MergeSorterColumnation<(B::Key, B::Val), B::Time, B::R>,
     lower: Antichain<B::Time>,
     frontier: Antichain<B::Time>,
-    phantom: ::std::marker::PhantomData<B>,
+    phantom: PhantomData<B>,
 }
 
 impl<B: Batch> Batcher<B> for ColumnatedMergeBatcher<B>
@@ -36,7 +37,7 @@ impl<B: Batch> Batcher<B> for ColumnatedMergeBatcher<B>
             sorter: MergeSorterColumnation::new(),
             frontier: Antichain::new(),
             lower: Antichain::from_elem(<B::Time as timely::progress::Timestamp>::minimum()),
-            phantom: ::std::marker::PhantomData,
+            phantom: PhantomData,
         }
     }
 
@@ -77,11 +78,9 @@ impl<B: Batch> Batcher<B> for ColumnatedMergeBatcher<B>
             for datum @ ((key, val), time, diff) in &buffer[..] {
                 if upper.less_equal(time) {
                     self.frontier.insert(time.clone());
-                    if keep.len() == keep.capacity() {
-                        if keep.len() > 0 {
-                            kept.push(keep);
-                            keep = self.sorter.empty();
-                        }
+                    if !keep.is_empty() && keep.len() == keep.capacity() {
+                        kept.push(keep);
+                        keep = self.sorter.empty();
                     }
                     keep.copy(datum);
                 }
@@ -99,14 +98,14 @@ impl<B: Batch> Batcher<B> for ColumnatedMergeBatcher<B>
         }
 
         // Finish the kept data.
-        if keep.len() > 0 {
+        if !keep.is_empty() {
             kept.push(keep);
         }
-        if kept.len() > 0 {
+        if !kept.is_empty() {
             self.sorter.push_list(kept);
         }
 
-        // Drain buffers (fast reclaimation).
+        // Drain buffers (fast reclamation).
         self.sorter.clear_stash();
 
         let seal = builder.done(self.lower.clone(), upper.clone(), Antichain::from_elem(<B::Time as timely::progress::Timestamp>::minimum()));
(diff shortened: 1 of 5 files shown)


src/trace/implementations/ord.rs
```diff
@@ -163,11 +161,13 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
 
 /// An immutable collection of update tuples, from a contiguous interval of logical times.
 #[derive(Abomonation)]
-pub struct OrdValBatch<L: Layout> {
+pub struct OrdValBatch<L: Layout, C> {
```
frankmcsherry, line 164: I'm reading this, and .. while I can probably figure it out I have no clue what `C` is meant to be. Would you be willing to add some doc text about it? It's a bit awkward that it plays no role for some time, leaving no hints about what it is.

src/trace/mod.rs
```diff
@@ -253,7 +254,7 @@ where
     /// Values associated with keys.
     type Val;
     /// Timestamps associated with updates
-    type Time;
+    type Time: Lattice+Timestamp;
```
frankmcsherry, line 257: I'm not fundamentally opposed, but is there a reason this type needs a constraint but the others do not?

src/trace/implementations/ord.rs
frankmcsherry: I've read a bit more, and .. it seems like this is mostly a newtype hack? Like, `C` could be `bool` and `String` and line up with the `Vec` and `TimelyStack` merge batchers, if the `impl`s said that. It feels a bit weird!

antiguru: I think we could make the merge batcher part of the layout and it would resolve all these issues. But it also feels odd because it's not really part of the layout... What I can't wrap my head around is how to get from an update container on the layout to a merge batcher implementation.

src/trace/mod.rs
antiguru: Removed, stray.

src/trace/implementations/ord.rs
antiguru: Nope, I can't see how to make it part of the layout. We don't know the `Batch` yet, and the batcher needs to be parametrized with `Self` a.k.a. the batch type :/

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+        TimelyStackQueue {
+            list,
+            head: 0,
+        }
+    }
+
+    fn done(mut self) -> TimelyStack<T> {
+        self.list.clear();
+        self.list
+    }
```
frankmcsherry, line 150: I expected to see this logic in `recycle()` instead. Is there a reason to have it here rather than there? Sorta feels like you could have the recycler clean things up, and then remove this method. Maybe you need something to recover the stack from the queue, but .. idk maybe it's good! It's not that it doesn't work, just trying to keep my head clear about whether we are laying booby traps by e.g. having `recycle` panic if you pass it a non-empty thing, rather than doing the `clear` there.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Remove all elements from the stash.
+    fn clear_stash(&mut self) {
+        self.stash.clear();
+    }
+
+    /// Insert an empty buffer into the stash. Panics if the buffer is not empty.
+    fn recycle(&mut self, buffer: TimelyStack<(D, T, R)>) {
+        assert!(buffer.is_empty());
```
frankmcsherry, line 203: Mentioned elsewhere, but this might be a time to call `buffer.clear()` instead, as 1. it should be cheap if empty, 2. we don't want to retain the regions if it is empty but not clear, and 3. avoids a panic booby trap.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+
+            if result.capacity() == result.len() {
+                output.push(result);
+                result = self.empty();
+            }
+
+            if head1.is_empty() {
+                let done1 = head1.done();
+                self.recycle(done1);
+                head1 = if list1.peek().is_some() { TimelyStackQueue::from(list1.next().unwrap()) } else { TimelyStackQueue::new() };
```
frankmcsherry, line 320: Another moment where it seems like it could be simpler, but maybe not for a specific reason!

frankmcsherry (approved):
I left several comments, only a few of which are "important" to chase down and most are either questions or stylistic nits. The main one I'd like to figure out is whether you want to double the capacity of `pending` to ensure you ship full containers. That seems relatively impactful. Several other discussion points that we can chase down here or 1:1.

frankmcsherry:
I have another question too: could the merger be for arbitrary `BatchContainer` types, or does it just work for `TimelyStack`? Is there some `BC::prefered_capacity()` method we might add to make it general?

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+    const fn buffer_size() -> usize {
+        let size = std::mem::size_of::<(D, T, R)>();
+        if size == 0 {
+            Self::BUFFER_SIZE_BYTES
+        } else if size <= Self::BUFFER_SIZE_BYTES {
+            Self::BUFFER_SIZE_BYTES / size
+        } else {
+            1
+        }
+    }
```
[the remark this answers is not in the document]
frankmcsherry: Also, and worth discussing: I think this size has roughly nothing to do with what `lg_alloc` will have to absorb? It indicates how large a `Vec<(D, T, R)>` will be, but the container will instead hold on to the spilled data, which can be arbitrarily larger (e.g. if `D` is `mz_repr::Row`), or arbitrarily smaller. It might be a good set-up for when you are able to back the `Vec` itself in `lg_alloc`, but .. is it really doing the right thing until then?

Seems not terrible ofc, but idk if we are sure that the backing allocations will be able to be paged out using this logic. Perhaps it's fine, in that it ensures that we take a large enough bite that the `Vec` would be that large, and if the alloc is smaller then we have at most 2x the `Vec` in memory (it plus a bit less in the region, in the worst case), which is decent insurance to provide (vs a smaller `Vec` which might have proportionally more in the region, and still not spill out).

antiguru:
> I have another question too: could the merger be for arbitrary `BatchContainer` types, or does it just work for `TimelyStack`? Is there some `BC::prefered_capacity()` method we might add to make it general?

I'd prefer a single general implementation, but it's not easy to get there:
* For the vector-based implementation, we can recycle buffers from the input and don't need to traverse the `pending` buffer. We could, but we first would need to do the measurements to see what impact it might have.
* For Columnation-based containers, we have a nice `copy_destructured` API, which allows us to avoid allocating a tuple just because we have no alternative way to represent its parts. We don't have this API for `Vec`, so we'd need to invent something.
* The rest seems to be mostly typing work, so not very difficult, and it might create some clarity around the `C` parameter to `OrdValBatch`.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+                    }
+                    keep.copy(datum);
+                }
+                else {
+                    builder.push((key.clone(), val.clone(), time.clone(), diff.clone()));
+                }
+                if let Some(last) = last {
+                    assert!(last <= datum);
+                }
+                last = Some(datum);
```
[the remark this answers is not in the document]
antiguru: Removed, it makes more sense to have this within `finish` and not here.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+        self.list
+    }
+
+    fn len(&self) -> usize { self.list.len() - self.head }
+
+    fn is_empty(&self) -> bool { self.head == self.list.len() }
+
+    /// Return an iterator over the remaining elements.
+    fn iter(&self) -> impl Iterator<Item=&T> + Clone + ExactSizeIterator {
+        self.list.iter().skip(self.head)
```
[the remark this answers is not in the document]
antiguru: It should use the `iter` implementation from `Deref<[T]>`, but made it explicit to avoid regressions.

src/trace/implementations/merge_batcher_col.rs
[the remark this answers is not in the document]
antiguru: I added a `pending_buffer_size` function that returns `2*Self::buffer_size()` so that pending contains enough data to fill a whole buffer. I need to test the performance impact of this, but it seems like the right change to do.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+    }
+
+    fn push(&mut self, batch: &mut Vec<(D, T, R)>) {
+        if self.pending.capacity() == 0 {
+            self.pending.reserve(Self::buffer_size());
+        }
+
+        while !batch.is_empty() {
+            self.pending.extend(batch.drain(..std::cmp::min(batch.len(), self.pending.capacity() - self.pending.len())));
```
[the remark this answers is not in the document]
antiguru: Updated the resizing logic above and added a comment to clarify our invariants on the capacity.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+    /// Move all elements in `pending` into `queue`. The data in `pending` must be compacted and
+    /// sorted.
+    fn flush_pending(&mut self) {
+        if !self.pending.is_empty() {
+            let mut stack = self.empty();
+            stack.reserve_items(self.pending.iter());
+            for tuple in self.pending.iter() {
+                stack.copy(tuple);
+            }
+            self.pending.clear();
```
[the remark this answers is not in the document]
antiguru: Done.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        // TODO: `list1` and `list2` get dropped; would be better to reuse?
+        let mut output = Vec::with_capacity(list1.len() + list2.len());
+        let mut result = self.empty();
+
+        let mut list1 = list1.into_iter().peekable();
+        let mut list2 = list2.into_iter().peekable();
+
+        let mut head1 = if list1.peek().is_some() { TimelyStackQueue::from(list1.next().unwrap()) } else { TimelyStackQueue::new() };
+        let mut head2 = if list2.peek().is_some() { TimelyStackQueue::from(list2.next().unwrap()) } else { TimelyStackQueue::new() };
```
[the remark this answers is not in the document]
antiguru: Simplified, this was a leftover from the old merge batcher implementation.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+struct TimelyStackQueue<T: Columnation> {
+    list: TimelyStack<T>,
+    head: usize,
+}
+
+impl<T: Columnation + 'static> TimelyStackQueue<T> {
+
+    fn new() -> Self { TimelyStackQueue::from(Default::default()) }
```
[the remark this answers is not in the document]
antiguru: We check on recycling whether the capacity is the default capacity and only then retain the allocation. Otherwise, we never write to a `TimelyStackQueue`.

src/trace/implementations/merge_batcher_col.rs
antiguru: Consolidated in `recycle()`, which now doesn't panic anymore.

src/trace/implementations/merge_batcher_col.rs
antiguru: This now check that the capacity matches the expected capacity, clears and stashes, or drops it otherwise. No more panics.

src/trace/implementations/merge_batcher_col.rs
antiguru: Yup, simplified.

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn push(&mut self, batch: &mut Vec<(D, T, R)>) {
+        if self.pending.capacity() == 0 {
+            self.pending.reserve(Self::buffer_size());
+        }
+
+        while !batch.is_empty() {
+            self.pending.extend(batch.drain(..std::cmp::min(batch.len(), self.pending.capacity() - self.pending.len())));
+            if self.pending.len() == self.pending.capacity() {
+                crate::consolidation::consolidate_updates(&mut self.pending);
+                if self.pending.len() > self.pending.capacity() / 2 {
```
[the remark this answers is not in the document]
antiguru: Changed it such that pending has twice the capacity of `TimelyStack`. I'll do some testing to figure out how it might affect performance.

frankmcsherry:
Looks good, and let's merge!

PR #419, opened 2023-11-19
Implement `OrdValBatch` without `retain_from`

frankmcsherry:
This PR provides a re-implementation of `OrdValBatch` with a few properties:
1. it is more direct than the existing implementation, based off of trie layers,
2. it only writes advanced and consolidated updates, rather than rewriting them later,
3. it is amenable to container implementations for `Vec<(Time, Diff)>`,
4. it is amenable to compact `Vec<(Time, Diff)>` representations (RLE).

There might be other things to like about it. In the fullness of time it would mean we could remove the `trace/layers` module, because the direct implementations aren't much more complicated, and in some ways are much simpler because of their directness.

This PR passes tests, but it probably wants a fair bit of exercise to see if it exactly tracks the existing implementations.

cc: @antiguru

### DIFF
--- a/src/trace/implementations/ord_neu.rs
+++ b/src/trace/implementations/ord_neu.rs
@@ -10,108 +10,17 @@
 
 use std::rc::Rc;
 
-use timely::container::columnation::TimelyStack;
-use timely::container::columnation::Columnation;
-
-use ::difference::Semigroup;
-use lattice::Lattice;
-
-use trace::layers::BatchContainer;
-use trace::layers::ordered::OrdOffset;
 use trace::implementations::spine_fueled::Spine;
 
+use super::{Update, Layout, Vector, TStack};
+
 use self::val_batch::{OrdValBatch};
-
-/// A type that names constituent update types.
-pub trait Update {
-    /// Key by which data are grouped.
-    type Key: Ord+Clone;
-    /// Values associated with the key.
-    type Val: Ord+Clone;
-    /// Time at which updates occur.
-    type Time: Ord+Lattice+timely::progress::Timestamp+Clone;
-    /// Way in which updates occur.
-    type Diff: Semigroup+Clone;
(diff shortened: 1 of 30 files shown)


src/trace/implementations/ord_neu.rs
```diff
@@ -0,0 +1,565 @@
...
+        fn cursor(&self) -> Self::Cursor { 
+            OrdValCursor {
+                key_cursor: 0,
+                val_cursor: 0,
+                phantom: std::marker::PhantomData,
+            }
+        }
+        fn len(&self) -> usize { 
+            // Normally this would be `self.updates.len()`, but we have a clever compact encoding.
+            // Perhaps we should count such exceptions to the side, to provide a correct accounting.
```
frankmcsherry, line 185: No clever compact encoding yet, so ignore this comment.

frankmcsherry:
The most recent commit reorganizes the two spine implementations, and re-exports `ord.rs`'s spines (the old ones) as `ValSpine` and `KeySpine` from their shared module root. All direct uses of e.g. `OrdValSpine` are replaced with `ValSpine`. The notable exception is `columnation.rs` which is meant to exercise columnation and non-columnation spines, so I left it pointing at the specific ones.

frankmcsherry:
Some light performance investigation suggests that this is not universally worse than the `ord.rs` spine. Running the `bfs` example with n=10^6, m=2x10^6, batches of size 1000, 1000 rounds completes in
```
ord    42.236413s
neu    39.531365375s
```
Running with n=10^8 and m=2x10^8, it takes both of them roughly the same time to load the data
```
ord    38.825862291s
neu    37.227272667s
```
Anecdotally through Activity Monitor, it appeared that `ord` went up to about 7GB where `neu` stayed around `3GB`. That was very unscientific on my part, though.

PR #420, opened 2023-11-20
Quality of life updates: non-optional compaction, and `Builder::copy`.

frankmcsherry:
Two changes; I thought there would be more but stopping here for now:
1. Compaction used to be an optional argument for merging. However, we never use that and it's something of an antipattern (in ye olde days, we only compacted the largest batch in merges).
2. Added a `copy` method to `Builder` allowing for non-owned presentation of tuples. This is especially useful when the destination container does not need to take ownership. Unfortunately we do not retrofit the `TupleBulider` trait used by trie layers, as they conceal their `Item` type structure well enough to not support e.g. `(&Key, &Val)` in place of `&(Key, Val)`.


antiguru (approved):
Nice!

PR #421, opened 2023-11-20
Encode singleton repetition update cleverly

frankmcsherry:
This PR introduces a clever optimization for representing in a batch any `(Key, Val)` pairs with a singleton `(Time, Diff)` that happens to exactly match the previous `(Time, Diff)`. This happens often in e.g. snapshot batches, where often the time is "zero" (or equivalent modulo frontiers) and the diff is "one".

Conventionally for each `(Key, Val)` we record an offset that indicates the upper bound of the corresponding `(Time, Diff)` entries, starting from the previous offset: the upper bound of the prior keyvals list of timediffs. Conventionally, each recorded offset must be strictly greater than the offset that precedes it, because we simply don't record absent updates.

We take advantage of this to signal a repeated timediff by repeating the offset. Essentially, if two updates `(lower, upper)` are equal, the range that should be used is instead `(lower - 1, upper)`, picking up the single previous timediff. We need to be careful to read out the right data, also to encode the data correctly, and also to report the total number of updates correctly (it is no longer `updates.len()`).

When running 
```
cargo run --release --example bfs -- 100000000 200000000 1000 0 potato
```
this change improves the limiting memory use (at the end of the execution, with all data in batches) from ~2.75GB to ~1.50GB. This program uses `u32` keys and values, which means that the times and diffs are substantial fat to cut off.

### DIFF
--- a/src/trace/implementations/ord_neu.rs
+++ b/src/trace/implementations/ord_neu.rs
@@ -69,7 +69,15 @@ mod val_batch {
         }
         /// Lower and upper bounds in `self.updates` corresponding to the value at `index`.
         fn updates_for_value(&self, index: usize) -> (usize, usize) {
-            (self.vals_offs[index].try_into().ok().unwrap(), self.vals_offs[index+1].try_into().ok().unwrap())
+            let mut lower = self.vals_offs[index].try_into().ok().unwrap();
+            let upper = self.vals_offs[index+1].try_into().ok().unwrap();
+            // We use equal lower and upper to encode "singleton update; just before here".
+            // It should only apply when there is a prior element, so `lower` should be greater than zero.
+            if lower == upper {
+                assert!(lower > 0);
+                lower -= 1;
+            }
+            (lower, upper)
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/trace/implementations/ord_neu.rs
```diff
@@ -425,6 +453,41 @@ mod val_batch {
...
+                assert!(self.singleton.is_none());
+                self.singleton = Some((time, diff));
+            }
+            else {
+                // If we have pushed a single element, we need to copy it out to meet this one.
+                if let Some(time_diff) = self.singleton.take() {
+                    self.result.updates.push(time_diff);
+                }
+                self.result.updates.push((time, diff));
+                self.singleton = None;
```
antiguru, line 488: No need to assign `None` as you `take()` above.

antiguru (approved):
Looks good! The logic is subtle; but I think it does the right thing: Pushing the same (key, value) into the builder actually spells out the prior singleton, and in all other cases, we take it and increment the singleton counter.

src/trace/implementations/ord_neu.rs
frankmcsherry: Good point.

PR #425, opened 2023-11-23
Only retain two allocations

antiguru:
This fixes retaining memory as large as the output when sealing the merge batcher for columnation.


frankmcsherry:
This makes sense to me, after conversation with you!

antiguru:
Confirming that it fixes the issue observed in https://github.com/MaterializeInc/materialize/issues/23389.

PR #427, opened 2023-11-23
Further trait reorganization

frankmcsherry:
This PR moves `BatchContainer` and `OrdOffset` out of the `layers` module which is specific to Trie implementations, and in to the `implementations` module, which seems more appropriate.

It also experiments a bit with the `BatchContainer` trait, allowing `Item` to be unsized, and supporting `push` and `copy_slice` only when `T: ToOwned`. This is .. unfortunately a restriction rather than a pure generalization; to implement `ToOwned` you have to be able to clone yourself, even though we don't need this. 

Up for discussion, but I think based on the vibes around containers, and their non-ownership-providing nature, this might be ok. 

cc: @antiguru


frankmcsherry:
Oh oops, it also adds a `SliceContainer` which shows off the new `BatchContainer` by having an `Item = [B]` but accepting `Vec<B>` in `push` and `copy_slice`, and storing only one `Vec<B>` and a whole bunch of offsets into it (as many as you introduce, plus one).

src/trace/implementations/mod.rs
```diff
@@ -167,3 +168,251 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
...
+                self.copy(other.index(index));
+            }
+        }
+        fn with_capacity(size: usize) -> Self {
+            Self {
+                offsets: Vec::with_capacity(size),
+                inner: Vec::with_capacity(size),
+            }
+        }
+        fn reserve(&mut self, _additional: usize) {
```
antiguru, line 391: Note to @antiguru: we could have a slightly better implementation here to pre-size the local part of self, but I can't recall the available apis out of my head.

src/trace/implementations/mod.rs
```diff
@@ -167,3 +168,251 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
...
+        fn index(&self, index: usize) -> &Self::Item {
+            &self[index]
+        }
+        fn len(&self) -> usize {
+            self[..].len()
+        }
+    }
+
+    // The `ToOwned` requirement exists to satisfy `self.reserve_items`, who must for now
+    // be presented with the actual contained type, rather than a type that borrows into it.
```
antiguru, line 306: We should revisit the type requirements on region and timely stack!

antiguru (approved):
Looks fine!

src/trace/implementations/mod.rs
frankmcsherry: Right! I'm not sure if it really helps here, but e.g. if I have a bunch of `Vec<u8>` and you are able to only maintain `[u8]` shaped data, that could help. Or we could conclude that this type does not do that, and not be too stressed! :D

PR #429, opened 2023-11-25
DNM: Revisiting ownership of key and value types

frankmcsherry:
DO NOT MERGE: This is a work in progress, and may ultimately fail.

This PR is an exploration of what happens if you want your key type `K` to primarily satisfy the constraint `K: ToOwned + ?Sized`. What?? The intent is that you might like to support key types like `str`, `[T]`, and other unsized types that can be observed behind a `&` reference.

The motivation is that we have containers that can represent `str` and `[T]` more efficiently than by storing actual `String` and `Vec<T>` objects, each of which have some 24 bytes of overhead. However, someone needs to do the translation between `[T]` and `Vec<T>`. This is where the `ToOwned` trait comes in: it allows `str` to express that it can be converted to an owned `String`, and that `String: Borrow<str>`. Pretty handy stuff!

Unfortunately, it's quite a lot of typing to convince Rust that this all works out. Fortunately, once Rust is convinced, I'm pretty sure too. But .. it takes some work getting there, and I'm not sure we've sorted everything out. This PR has only touched `join` and `arrange` operators so far, not `reduce`, and it commented out the entire `implementations/layers/` module as irredeemable (it can perhaps be redeemed, but with a bunch of work).


src/operators/arrange/arrangement.rs
```diff
@@ -488,54 +489,30 @@ where
...
-        V: ExchangeData,
-        R: ExchangeData,
-        Tr: Trace + TraceReader<Time=G::Timestamp> + 'static, Tr::Batch: Batch,
-        Tr::Batcher: Batcher<Item = ((K,V),G::Timestamp,R), Time = G::Timestamp>,
-        Tr::Builder: Builder<Item = ((K,V),G::Timestamp,R), Time = G::Timestamp, Output = Tr::Batch>,
-    {
-        let exchange = Exchange::new(move |update: &((K,V),G::Timestamp,R)| (update.0).0.hashed().into());
-        self.arrange_core(exchange, name)
-    }
-
```
frankmcsherry, line 531: These were redundant with the default implementations of the trait, so I removed them!

src/operators/arrange/arrangement.rs
```diff
@@ -690,30 +667,37 @@ where
...
-        P: ParallelizationContract<G::Timestamp, ((K,()),G::Timestamp,R)>,
-        Tr: Trace+TraceReader<Time=G::Timestamp>+'static,
-        Tr::Batch: Batch,
-        Tr::Batcher: Batcher<Item = ((K,()),G::Timestamp,R), Time = G::Timestamp>,
-        Tr::Builder: Builder<Item = ((K,()),G::Timestamp,R), Time = G::Timestamp, Output = Tr::Batch>,
-    {
-        self.map(|k| (k, ()))
-            .arrange_core(pact, name)
-    }
-}
```
frankmcsherry, line 708: This is too bad, but roughly there is a constraint satisfaction issue that Rust can't seem to determine that `(_, ())` can't unify with `_`, which it was previously able to determine for `_ = K`, but cannot when `_ = K::Owned`. It's possible that I could fix something here, but one can work around this by changing the call to `arrange` on non-pair data into `arrange_by_self`, which is the more explicit way to arrange that way.

src/trace/implementations/mod.rs
```diff
@@ -45,25 +45,30 @@ pub(crate) mod merge_batcher_col;
...
 use lattice::Lattice;
 use difference::Semigroup;
 
 /// A type that names constituent update types.
 pub trait Update {
+    /// We will be able to read out references to this type, and must supply `Key::Owned` as input.
+    type Key: Ord + ToOwned<Owned = Self::KeyOwned> + ?Sized;
     /// Key by which data are grouped.
-    type Key: Ord+Clone;
+    type KeyOwned: Ord+Clone + Borrow<Self::KeyOwned>;
```
frankmcsherry, line 71: This type is present here because it seems relatively hard to drive type inference off of `Key::Owned`, and in particular expressing constraints on that type (e.g. `Ord + Clone`). You could put a `where` clause with these constraints, but that just turned in to a where clause that all users of `Update` need to repeat as part of their constraints on `U: Update` (that it's key's offset implements certain traits).

Perhaps this can be finessed; it's certainly gross having to state the type twice, once explicitly and once as `Key::Owned`.

src/trace/implementations/mod.rs
```diff
@@ -80,15 +85,31 @@ where
     R: Semigroup+Clone,
 {
     type Key = K;
+    type KeyOwned = K;
     type Val = V;
     type Time = T;
     type Diff = R;
 }
 
+impl<K,V,T,R> Update for Box<((K, V), T, R)>
```
frankmcsherry, line 94: This `Box<_>` thing is just a random type. I couldn't get `[_]` to work because the unsizedness infected everything. Also, `(([K], V), T, D)` was just very upsetting to Rust; it never got off the ground.

src/trace/implementations/rhh.rs
```diff
@@ -742,7 +758,7 @@ mod val_batch {
         fn push(&mut self, ((key, val), time, diff): Self::Item) {
 
             // Perhaps this is a continuation of an already received key.
-            if self.result.keys.last() == Some(&key) {
+            if self.result.keys.last() == Some(key.borrow()) {
```
frankmcsherry, line 761: Here is one of the first examples of actually having to relate the types. I'm not even sure we use any methods of `ToOwned` anywhere else!

frankmcsherry:
The most recent commit adds support for `Val` being unsized, and only implementing `ToOwned`.

But also exciting, @antiguru, are the `PreferredContainer` trait and the `Preferred<K,V,T,R>` struct which implements `Update` and `Layout` and uses the containers preferred by the types (e.g. `Vec<String>` for `String`, and `SliceContainer<u8>` for `[u8]`). You can see an example of it in `examples/spines.rs`

frankmcsherry:
The most recent commit relaxes `Arrange` to speak of input types, rather than potentially unsized arrangement types. This .. makes sense because the `Arrange` trait itself has no opinions about the trace types to be used, which show up only in the `fn arrange<Tr>` methods themselves. At that point, constraints on the traces are introduced, but only in the form of what inputs are required. All constraints on output types, e.g. `[u8]` or potentially GATs, are absent from this trait, its methods, and its implementations (except that the timestamps need to line up).

frankmcsherry:
The most recent commit looks large, but it's almost all whitespace due to changing a tab indent level. They introduce functions `join_trace` and `reduce_trace` which are not part of a trait, and are generic in trace rather than key, val, etc. This is to present a minimally restrictive form of the functionality, partly to ensure we are able to write this only as a function of traces, rather than .. weird related types that might need owned and borrowed variants, etc. Nope! Or, mostly nope!

src/operators/arrange/arrangement.rs
frankmcsherry: This got fixed!

src/trace/implementations/merge_batcher.rs
```diff
@@ -12,14 +12,14 @@ use trace::implementations::Update;
 
 /// Creates batches from unordered tuples.
 pub struct MergeBatcher<U: Update> {
```
frankmcsherry, line 14: Perhaps this struct should just be parameterized by `K, V, T, D`?

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -14,25 +14,25 @@ use trace::implementations::Update;
 /// Creates batches from unordered tuples.
 pub struct ColumnatedMergeBatcher<U: Update>
```
frankmcsherry, line 15: Perhaps this struct should just be parameterized by `K, V, T, D`?

src/trace/implementations/rhh.rs
```diff
@@ -97,7 +98,8 @@ mod val_batch {
     #[derive(Abomonation, Debug)]
     pub struct RhhValStorage<L: Layout> 
     where 
-        <L::Target as Update>::Key: Default + HashOrdered
+        <L::Target as Update>::Key: HashOrdered,
+        <L::Target as Update>::KeyOwned: Default + HashOrdered,
```
frankmcsherry, line 102: This probably shouldn't be a requirement of the structure itself, but rather the implementations that need to push data in (`Default` so we can pad locations, and `HashOrdered` for actual owned insertions).

src/trace/implementations/merge_batcher_col.rs
frankmcsherry: Will address in follow-up work.

PR #430, opened 2023-11-25
Avoid reallocations in first `keep`, bound stash size

antiguru:
A follow-up to #425. Retain two elements instead of three in the stash. Avoid doubling reallocations in the first `keep` object.

The second change is up for discussion: If we expect few (more than zero) elements to be kept, it's worse to use a complete buffer, but the cost to reallocate should quickly outweigh the const of allocating a large buffer (or, reusing one for that matter.) This also changes if we already have a buffer in the stash, where it would be clearly beneficial to use that allocation instead of getting a new one.


frankmcsherry:
These changes seem good to me. Not sure what is up with the error, but modulo that looks good to me.

antiguru:
I should have `cargo check`ed before pushing :) Fixed!

PR #431, opened 2023-11-27
Remove batcher dependence on updates

frankmcsherry:
This PR removes both batchers' dependence on `U: Update`, replacing the dependence with the vanilla types it would extract from the update. This generally simplifies things when we know explicit types, and is a bit more confusing when we baked a more complicated rule into the update. We may need to re-jigger in the future, but I don't think we need to make these types update-aware.


antiguru (approved):
Looks good!

PR #433, opened 2023-11-27
Fix lookup_map compile error

antiguru:
It doesn't make any effort to relax the trait bounds in dogs^3, if possible at all.


frankmcsherry (approved):
Looks great; I'll follow up pondering about whether these are the right constraints, but this isn't an operator we use much, so .. probably all good! :D

PR #437, opened 2023-11-28
Replace `OrdOffset` with `OffsetList`

frankmcsherry:
The `OrdOffset` trait allowed us to slot in `u32` or `u64` in for `usize` for offsets, in order to save the RAMs. That isn't terrible, but we can remove the complexity, and the potential surprise if we go too large, with the `OffsetList` struct that generally behaves as a `Vec<u32>` until we reach a point at which the entries exceed its size, at which point it behaves like a not-great `Vec<u64>`.

### DIFF
--- a/src/trace/implementations/mod.rs
+++ b/src/trace/implementations/mod.rs
@@ -53,8 +53,6 @@ pub mod rhh;
 pub use self::ord::OrdValSpine as ValSpine;
 pub use self::ord::OrdKeySpine as KeySpine;
 
-use std::ops::{Add, Sub};
-use std::convert::{TryInto, TryFrom};
 use std::borrow::{Borrow, ToOwned};
 
 use timely::container::columnation::{Columnation, TimelyStack};
@@ -96,10 +94,6 @@ where
 pub trait Layout {
     /// The represented update.
     type Target: Update + ?Sized;
-    /// Offsets to use from keys into vals.
-    type KeyOffset: OrdOffset;
-    /// Offsets to use from vals into updates.
-    type ValOffset: OrdOffset;
     /// Container for update keys.
     type KeyContainer:
         RetainFrom<<Self::Target as Update>::Key>+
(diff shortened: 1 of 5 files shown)


src/trace/implementations/mod.rs
```diff
@@ -239,16 +225,90 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
...
+        }
+        else {
+            self.chonk[index - self.smol.len()].try_into().unwrap()
+        }
+    }
+    /// Set the offset at location index.
+    ///
+    /// Complicated if `offset` does not fit into `self.smol`.
+    pub fn set(&mut self, index: usize, offset: usize) {
+        if index < self.smol.len() {
```
antiguru, line 267: Maybe I'm misunderstanding, does this imply that `set` must be called with increasing offsets? Otherwise, we might alternate between putting things in smol and chonk. I think this is true the way we use it, but maybe we should change the condition to only take the first branch if `self.chunk.is_empty()`?

src/trace/implementations/mod.rs
```diff
@@ -239,16 +225,90 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
...
-where
-    O: Copy + PartialEq + Add<Output=Self> + Sub<Output=Self> + TryFrom<usize> + TryInto<usize>,
-{}
+/// A list of unsigned integers that uses `u32` elements as long as they are small enough, and switches to `u64` once they are not.
+#[derive(Eq, PartialEq, Ord, PartialOrd, Clone, Debug, Abomonation)]
+pub struct OffsetList {
+    /// Offsets that fit within a `u32`.
+    smol: Vec<u32>,
+    /// Offsets that either do not fit in a `u32`, or are inserted after some offset that did not fit.
+    chonk: Vec<u64>,
```
antiguru, line 236: What about extending this to `u16`? It would save some bytes for batches with less then 64k elements, which should be quite frequent.

antiguru (approved):
Nice!

src/trace/implementations/mod.rs
```diff
@@ -239,16 +225,90 @@ impl<T: Columnation> RetainFrom<T> for TimelyStack<T> {
...
+                self.migrate_to_chonk_from(index);
+                self.chonk[index - self.smol.len()] = offset.try_into().unwrap();
+            }
+        }
+        else {
+            self.chonk[index - self.smol.len()] = offset.try_into().unwrap();
+        }
+    }
+    fn migrate_to_chonk_from(&self, _index: usize) {
+        unimplemented!()
```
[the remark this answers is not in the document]
frankmcsherry: This was implemented!

src/trace/implementations/mod.rs
frankmcsherry: Discussed in person; we agree that this is great.

src/trace/implementations/mod.rs
frankmcsherry: I'm worried that the added pain when you use things that don't fit in a `u32` feels more acute for things that don't fit in a `u16`. Maybe adding this makes sense but it's now shaving off just two bytes, and .. I'm up for it if we don't care about the CPU cost, but someone using DD probably does (more than a hypothetical 4B+ user).

PR #440, opened 2023-12-01
Make contents of OffsetList pub

antiguru:
Make the contents of `OffsetList` pub. Although it's strange to have this as part of the public interface, it'll be useful to surface its size clients interested in it.


frankmcsherry:
Looks great!

PR #441, opened 2023-12-06
Huffman container example

frankmcsherry:
An example container that uses Huffman coding to store `[B]` types instead as `[u8]` with a side map to elements `B` based on their frequency. This container may or may not be directly useful, but it is meant to demonstrate some of the capabilities of Arrangement GATs.

### DIFF
--- a/src/trace/implementations/huffman_container.rs
+++ b/src/trace/implementations/huffman_container.rs
@@ -1,0 +1,569 @@
+//! A slice container that Huffman encodes its contents.
+
+use std::collections::BTreeMap;
+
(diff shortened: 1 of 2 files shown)


src/trace/implementations/huffman_container.rs
```diff
@@ -0,0 +1,569 @@
...
+                }
+            }
+
+            Huffman { 
+                encode,
+                decode,
+            }
+        }
+
+        /// Inserts a symbol, and
```
antiguru, line 381: Words missing

PR #446, opened 2023-12-09
Consolidate without a trace

antiguru:
Adds `consolidate_pact` that consolidates without creating a trace, and pivot `consolidate_named` to it.

### DIFF
--- a/src/operators/consolidate.rs
+++ b/src/operators/consolidate.rs
@@ -240,11 +240,13 @@ impl<'a, K: Data, V: Data, T: Timestamp, R: Data> Builder for ConsolidateBuilder
     type Output = ();
 
     fn new() -> Self {
-        unimplemented!()
+        // We never construct self using `new`.
+        unreachable!()
     }
 
     fn with_capacity(_keys: usize, _vals: usize, _upds: usize) -> Self {
-        unimplemented!()
+        // We never construct self using `with_capacity`.
+        unreachable!()
     }
 
     fn push(&mut self, element: Self::Item) {
--- a/src/operators/consolidate.rs
+++ b/src/operators/consolidate.rs
@@ -107,3 +107,157 @@ where
             .as_collection()
     }
 }
+
+pub mod neu {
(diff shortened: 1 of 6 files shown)


src/trace/implementations/merge_batcher_col.rs
```diff
@@ -62,7 +63,7 @@ where
     // which we call `lower`, by assumption that after sealing a batcher we receive no more
     // updates with times not greater or equal to `upper`.
     #[inline]
-    fn seal<B: Builder<Item=Self::Item, Time=Self::Time>>(&mut self, upper: Antichain<T>) -> B::Output {
+    fn seal<B: Builder<Item=Self::Item, Time=Self::Time>, C: FnOnce(usize, usize, usize) -> B>(&mut self, upper: AntichainRef<T>, constructor: C) -> B::Output {
```
antiguru, line 66: I'm not particularly happy about this `C` parameter, but it seemed easy to implement. Open to suggestions.

src/operators/consolidate.rs
```diff
@@ -106,3 +106,157 @@ where
...
+                    let mut prev_frontier = Antichain::from_elem(<G::Timestamp as Timestamp>::minimum());
+
+
+                    move |input, output| {
+                        input.for_each(|cap, data| {
+                            capabilities.insert(cap.retain());
+                            batcher.push_batch(data);
+                        });
+
+                        if prev_frontier.borrow() != input.frontier().frontier() {
```
antiguru, line 182: I stole the capability management logic from arrange_core, so it might check out?

src/trace/implementations/merge_batcher_col.rs
frankmcsherry: Oh interesting. This may be another pattern to explore in the "how do we hook up various interfaces" saga. :D

antiguru:
We implemented this as part of Materialize (https://github.com/MaterializeInc/materialize/pull/24377). Closing for the moment as there is a work-around.

PR #449, opened 2023-12-18
Let Layout specify OffsetContainer

antiguru:
This enhances `Layout` to carry a type `OffsetContainer`. Batch implementations should use this over directly using a specific implementation.


frankmcsherry:
I pushed up a small edit that allows more general `ReadItem` types for the offset list containers. Looks good, though.

PR #450, opened 2023-12-20
Fix off-by-one capacity error when sizing target containers

antiguru:


frankmcsherry (commented):
Ah, makes sense but could be we instead change the initial values to be `1` instead of `0`?

antiguru:
Yes; we could change the default to 1, but then we'd allocate a container of size 1 even if there's no data to be pushed? Might add a cost where we create empty batches.

frankmcsherry:
Makes sense!

PR #452, opened 2024-01-03
Introduce container for default values

frankmcsherry:
This PR introduces a container that effectively represents a run of default values, and spills to a back-up container if non-default values are discovered. This is meant to be a MVP example of a non-trivial container that relies on GATs, but also has the potential to bring `OrdVal` and `OrdKey` spines closer together, at some CPU cost.

### DIFF
--- a/src/trace/implementations/option_container.rs
+++ b/src/trace/implementations/option_container.rs
@@ -1,0 +1,151 @@
+//! A container optimized for identical contents.
+
+use crate::trace::cursor::MyTrait;
+use crate::trace::implementations::BatchContainer;
+
+/// A container that effectively represents default values.
+///
+/// This container is meant to be a minimal non-trivial container,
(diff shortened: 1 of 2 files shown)


src/trace/implementations/option_container.rs
```diff
@@ -0,0 +1,151 @@
...
+        else {
+            OptionWrapper { inner: Some(self.container.index(index - self.defaults))}
+        }
+    }
+    fn len(&self) -> usize {
+        self.container.len() + self.defaults
+    }
+}
+
+/// A read wrapper that presents as
```
antiguru, line 71: Missing text

antiguru (approved):
Looks good! Thanks

frankmcsherry:
Thanks for the review (made the fixes!)

PR #454, opened 2024-01-07
Offset list counts zero offsets

antiguru:
This change allows the OffsetList to count the length of a zero prefix of all offsets without actually storing the zero elements. This avoids consuming memory in a frequently-occurring pattern where at least the first element is zero, or if the singleton update optimization kicks in, a large prefix of elements (up to all) are zero.


frankmcsherry (approved):
I think this all checks out; thanks!

PR #456, opened 2024-01-22
Base update capacity off updates instead of old capacity

antiguru:
This makes sure we never reallocate the update container. Without this, we can allocate a container that is too small to contain all updates that we might generate during merging.


antiguru:
Closing this because the PR makes a worse estimation to improve virtual memory. Virtual memory usually isn't a constrained resource, so I prefer not merging this PR.

PR #458, opened 2024-01-30
Canonicalize `PointStamp` representation

frankmcsherry:
`PointStamp` represents an infinite vector of `T` elements, extending a maintained `vector` with `T::minimum()`. This allows multiple representations of the same vector when the `vector` is allowed to end with `T::minimum()`, which was documented as acceptable. However, it does mean that e.g. the derived `Eq` implementation, among others, would report two logically identical pointstamps as different, and imo the position that there can be multiple unequal representations of the same state is untenable.

This PR removes that ability, short of direct manipulation by users, through an `enforce` method that is called by a constructor, and use of that constructor whenever we mint new `PointStamp`s. Demonstration uses that manipulate `vector` directly are followed by `enforce` calls.

We could go further and make the field private, and prevent direct manipulation and require clones (or at least `into_vec` and `from_vec` patterns). That might increase confidence even further.

cc: @antiguru

### DIFF
--- a/src/dynamic/pointstamp.rs
+++ b/src/dynamic/pointstamp.rs
@@ -17,7 +17,7 @@ use serde::{Deserialize, Serialize};
 /// A sequence of timestamps, partially ordered by the product order.
 ///
 /// Sequences of different lengths are compared as if extended indefinitely by `T::minimum()`.
-/// Sequences are not guaranteed to be "minimal", and may end with `T::minimum()` entries.
+/// Sequences are guaranteed to be "minimal", and may not end with `T::minimum()` entries.
 #[derive(
     Hash, Default, Clone, Eq, PartialEq, Ord, PartialOrd, Debug, Serialize, Deserialize, Abomonation,
 )]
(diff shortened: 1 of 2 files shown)


src/dynamic/pointstamp.rs
```diff
@@ -26,10 +26,20 @@ pub struct PointStamp<T> {
...
+    ///
+    /// This method will modify `vector` to ensure it does not end with `T::minimum()`.
     pub fn new(vector: Vec<T>) -> Self {
-        PointStamp { vector }
+        let mut result = PointStamp { vector };
+        result.enforce();
+        result
+    }
+    /// Enforces that `self` not end with `T::minimum()` by popping elements until it is true.
+    pub fn enforce(&mut self) {
```
antiguru, line 39: Does `enforce` need to be `pub`? It seems the only way to construct a `PointStamp` is through `new`, and all other places that would modify the vector go through this API? In this case, nobody should need to explicitly call `enforce` I believe -- this is not true if there is an API that modifies the pointstamp such that it doesn't have the canonical representation anymore.

Oh, I see: This is relevant as long as `vector` is pub. If that changes, we can probably make this method private.

antiguru (approved):
Thanks, looks good!

antiguru (approved):

src/dynamic/pointstamp.rs
frankmcsherry: Resolved as "`enforce` deleted".

PR #461, opened 2024-02-22
Change parameters to exert_logic from iterator to slice

antiguru:
This avoid the intermediate allocation for every invocation of exert_logic to create a boxed iterator. Instead, pass a slice that contains the same information previously revealed by the iterator. At the same time, wrap the logic in an option to only invoke it if it is set.

This is the source of around 80% of all allocations in mostly idle Materialize clusters. The lower two large blocks are related to this in the following image.

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,18 +1,43 @@
-name: test
-
-on: [push, pull_request]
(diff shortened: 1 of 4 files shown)


frankmcsherry:
Looks good to me!

PR #463, opened 2024-02-27
DNM: A merge batcher that gracefully handles non-ready data

antiguru:
This PR shows how to implement a merge batcher that is smart about reconsidering data that's in advance of the current frontier. It does the following things:
1. It extracts ready data from chains instead of merging all chains and then extracting data from the last remaining chain.
2. It separates canonicalization from the extraction operation, so it can be reused for inserts and extracts.
3. It memorizes a frontier per block, which allows for efficient frontier testing: If the extraction (upper) frontier is not less or equal to the block frontier, do not touch the block.

This should have the potential to reduce the amount of work for outstanding data from $O(n)$ where $n$ is the number of records in the merge batcher to $O(n/1024)$ by considering only the block itself, but not the data it contains.

I am sorry for the formatting noise which originates from copying this code from DD to Mz and back again :/

### DIFF
--- a/src/trace/implementations/merge_batcher.rs
+++ b/src/trace/implementations/merge_batcher.rs
@@ -55,10 +55,9 @@ where
     // in `upper`. All updates must have time greater or equal to the previously used `upper`,
     // which we call `lower`, by assumption that after sealing a batcher we receive no more
     // updates with times not greater or equal to `upper`.
-    #[inline]
     fn seal<B: Builder<Item=Self::Item, Time=Self::Time>>(&mut self, upper: Antichain<T>) -> B::Output {
         self.frontier.clear();
-        let merged = self.sorter.extract_into(upper.borrow(), &mut self.frontier);
+        let extracted = self.sorter.extract(upper.borrow(), &mut self.frontier);
 
         // Determine the number of distinct keys, values, and updates,
         // and form a builder pre-sized for these numbers.
@@ -67,7 +66,7 @@ where
             let mut vals = 0;
             let mut upds = 0;
             let mut prev_keyval = None;
-            for ((key, val), _time, _) in merged.iter().map(|t| t.iter()).flatten() {
+            for ((key, val), _time, _) in extracted.iter().flatten() {
                 if let Some((p_key, p_val)) = prev_keyval {
                     if p_key != key {
                         keys += 1;
@@ -85,9 +84,9 @@ where
             B::with_capacity(keys, vals, upds)
         };
 
-        for buffer in merged.into_iter() {
-            for datum in &buffer[..] {
-                builder.copy(datum);
+        for mut buffer in extracted {
+            for datum in buffer.drain(..) {
+                builder.push(datum);
             }
             // Recycling buffer.
             self.sorter.recycle(buffer);
@@ -112,9 +111,11 @@ where
 }
 
 struct MergeSorter<D, T, R> {
-    /// each power-of-two length list of allocations. Do not push/pop directly but use the corresponding functions.
+    /// each power-of-two length list of least times and allocations. Do not push/pop directly but use the corresponding functions.
     queue: Vec<Vec<(Antichain<T>, Vec<(D, T, R)>)>>,
+    /// Empty, recycled allocations. Use [`MergeSorter::emtpy`] to pop an allocation.
     stash: Vec<Vec<(D, T, R)>>,
+    /// Data that was pushed but not yet inserted into queue. Not necessarily sorted or compacted.
     pending: Vec<(D, T, R)>,
     logger: Option<Logger<DifferentialEvent, WorkerIdentifier>>,
     operator_id: usize,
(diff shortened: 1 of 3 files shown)


src/trace/implementations/merge_batcher.rs
```diff
@@ -179,81 +143,235 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
+        let mut keep = self.empty();
+        let mut keep_frontier = Antichain::new();
+        let mut ship = self.empty();
+        let mut ship_list = Vec::default();
+
+        self.account(self.queue.iter().flatten(), -1);
+
+        // Walk all chains, separate ready data from data to keep.
+        for mut chain in std::mem::take(&mut self.queue).drain(..) {
+            let mut block_list = Vec::default();
```
frankmcsherry, line 280: Said aloud: "could this be `ship_list`?". Maybe not, but the name `block_list` is not very specific in this context (many lists of blocks here).

src/trace/implementations/merge_batcher.rs
```diff
@@ -179,81 +143,235 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
+                    // All data is ready, push what we accumulated, stash whole block.
+                    if !ship.is_empty() {
+                        block_list.push((Antichain::new(), std::mem::replace(&mut ship, self.empty())));
+                    }
+                    block_list.push((Antichain::new(), block));
+                } else if any {
+                    // Iterate block, sorting items into ship and keep
+                    for datum in block.drain(..) {
+                        if upper.less_equal(&datum.1) {
+                            frontier.insert_ref(&datum.1);
```
frankmcsherry, line 298: This appears to be the only (?) place we update `frontier`, even though when we ship none of the updates (the `else` case below) we still want to reflect those times in the overall frontier.

src/trace/implementations/merge_batcher.rs
```diff
@@ -179,81 +143,235 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
+                    // Recycle leftovers
+                    self.recycle(block);
+                } else {
+                    // Keep the entire block.
+                    if !keep.is_empty() {
+                        keep_list.push((std::mem::take(&mut keep_frontier), std::mem::replace(&mut keep, self.empty())));
+                    }
+                    keep_list.push((block_frontier, block));
+                }
+            }
```
frankmcsherry, line 322: This is potentially a good moment, as we go, to perform the "adjacent blocks" compaction. That would allow us to return memory to `self.empty()` eagerly, and have it available as we loop. Waiting until `maintain()` is not wrong, but it seems like it may cause memory to spike during `extract_into` and return down only after `maintain()`.

src/trace/implementations/merge_batcher.rs
```diff
@@ -179,81 +143,235 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
+        if ship_list.len() > 0 {
+            self.maintain();
         }
 
-        if let Some(mut last) = self.queue_pop() {
-            ::std::mem::swap(&mut last, target);
+        while ship_list.len() > 1 {
+            let list1 = ship_list.pop().unwrap();
+            let list2 = ship_list.pop().unwrap();
+            ship_list.push(self.merge_by(list1, list2));
```
frankmcsherry, line 350: No reason these need to be in the same geometric order as the held updates. There is the opportunity to resort them, or to use a binary heap to continually merge the smallest chains. No strong feelings, because it seems unlikely to be adversarially selected.

src/trace/implementations/merge_batcher.rs
```diff
@@ -265,82 +383,89 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
-
-    /// Push a batch to `self.queue` and account size changes.
-    #[inline]
-    fn queue_push(&mut self, batch: Vec<Vec<(D, T, R)>>) {
-        self.account(batch.iter().map(Vec::len), 1);
-        self.queue.push(batch);
-    }
-
+impl<D, T, R> MergeSorter<D, T, R>
+{
```
frankmcsherry, line 455: ???

antiguru:
We figured that the approach taken in this PR changes what we report as the next lower bound of data to be extracted. As before, `seal` captures a lower frontier of all the times in the batcher, but its semantics are different. Previously, the reported frontier was accurate, i.e., there existed data at the reported frontier. Now, it's any lower bound, but there is no guarantee that it's accurate.

The reason for this is that in the past, `seal` would merge all data into a single chain, which means that each `(d, t)` pair appears at most once. After this change, each `(d, t)` pair can occur in all chains, which means that we can only compute a lower bound frontier of the uncompacted data, but not the precise lower frontier of the compacted data. We maintain a logarithmic amount of chains.

This can become a problem when all data cancels out, and can cause an unknown amount of additional work for the rest of the system, because it needs to maintain more capabilities and might need to ask for more data more times.

We don't have an immediate solution for this problem, but there are some options:
* We can accept this problem if we know that most data doesn't cancel, so the lower frontier should be mostly accurate. This requires that operators can handle hallucinated frontiers.
* We can limit the implementation to be only used for totally ordered times, and combining it with time-sorted chains. This would allow us to drain chains from the beginning while the consolidation would cancel out the contents.
* Longer term, we could form chains of totally ordered data within a partially ordered domain, where each chain can be consolidated efficiently. I favor this option least because implementation uncertainties.

antiguru:
Closing because of the issue described above.

PR #465, opened 2024-03-11
Fix testing mdbook

antiguru:
Fix testing of mdbook by cleaning the working directory before running the test.


frankmcsherry (approved):
No idea how this stuff works, but lgtm!

PR #471, opened 2024-04-09
WIP: Extract batcher input to assoc type, arrange_core freestanding

antiguru:


frankmcsherry (approved):
This seems great; thank you very much!

PR #473, opened 2024-04-13
Further trait clean-up

frankmcsherry:
Further clean-up around the `Batcher` and `Builder` trait interfaces. The `Item` associated types have been rebranded `Input` and `Output` depending on their use, and are less tightly bonded to being individual updates. Some additional constraints removed, dead code removed, stuff like that.


antiguru (approved):
Looks good!

PR #474, opened 2024-04-17
Merge batcher generic over containers

antiguru:
Merge batcher that's generic over input containers and internal chains, with specific implementations.

## Ideas

At the moment, a merge batcher receives a stream of vectors. It consolidates the input vectors, and inserts them into its queue structure. When sealing, it extracts ready data and presents it record-by-record to the builder. It inserts future updates into its queue.

This introduces several opportunities to introduce containers:
* The input the batcher: The stream of updates received from a timely operator. The batcher requires each batch to be consolidated, i.e., sorted and differences for same data accumulated. It can either take consolidated data, or consolidate it itself. We're missing an abstraction that allows the batcher to consolidate containers, and often containers are read-only, or we only get to look at the data.
  * We could sort by permuting a `Vec<usize>` and copying into a new container in sorted order.
  * Consolidating requires addition, which either requires storage or mutability, which is odd for read-only containers. Not sure what to do here.
* The internal queue structure. We need to traverse the elements in order, and compare on data and time, accumulating diffs as we go. This requires specific knowledge about items of a container, which motivates moving the bulk of the implementation behind a trait to avoid generic parameters.
* Ready data presented to the builder. We have an opportunity to present all data to the builder at once, which relieves it from incrementally building batches. It unlocks "looking ahead" in the input data.

### DIFF
--- a/src/trace/implementations/merge_batcher.rs
+++ b/src/trace/implementations/merge_batcher.rs
@@ -3,17 +3,17 @@
 use std::collections::VecDeque;
 
 use timely::communication::message::RefOrMut;
-use timely::{Container, PartialOrder};
 use timely::logging::WorkerIdentifier;
 use timely::logging_core::Logger;
-use timely::progress::{frontier::Antichain, Timestamp};
 use timely::progress::frontier::AntichainRef;
+use timely::progress::{frontier::Antichain, Timestamp};
+use timely::{Container, PartialOrder};
 
-use crate::Data;
 use crate::consolidation::consolidate_updates;
 use crate::difference::Semigroup;
 use crate::logging::{BatcherEvent, DifferentialEvent};
 use crate::trace::{Batcher, Builder};
+use crate::Data;
 
 /// Creates batches from unordered tuples.
 pub struct MergeBatcher<M, T>
@@ -40,7 +40,7 @@ where
 
 impl<M, T> Batcher for MergeBatcher<M, T>
 where
-    M: Merger<Time=T>,
+    M: Merger<Time = T>,
     T: Timestamp,
 {
     type Input = M::Input;
@@ -70,7 +70,7 @@ where
     // in `upper`. All updates must have time greater or equal to the previously used `upper`,
     // which we call `lower`, by assumption that after sealing a batcher we receive no more
     // updates with times not greater or equal to `upper`.
-    fn seal<B: Builder<Input=Self::Output, Time=Self::Time>>(&mut self, upper: Antichain<T>) -> B::Output {
+    fn seal<B: Builder<Input = Self::Output, Time = Self::Time>>(&mut self, upper: Antichain<T>) -> B::Output {
         // Finish
         let chain = self.merger.finish(&mut self.stash);
         if !chain.is_empty() {
@@ -110,9 +110,10 @@ where
         self.frontier.borrow()
     }
 }
+
 impl<M, T> MergeBatcher<M, T>
 where
-    M: Merger
+    M: Merger,
 {
     /// Insert a chain and maintain chain properties: Chains are geometrically sized and ordered
     /// by decreasing length.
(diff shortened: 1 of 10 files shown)


src/trace/implementations/merge_batcher.rs
```diff
@@ -1,169 +1,226 @@
...
 
-struct MergeSorter<D, T, R> {
-    /// each power-of-two length list of allocations. Do not push/pop directly but use the corresponding functions.
-    queue: Vec<Vec<Vec<(D, T, R)>>>,
-    stash: Vec<Vec<(D, T, R)>>,
-    logger: Option<Logger<DifferentialEvent, WorkerIdentifier>>,
-    operator_id: usize,
+    // merges two sorted input lists into one sorted output list.
+    fn merge_by(&mut self, list1: Vec<M::Batch>, list2: Vec<M::Batch>) -> Vec<M::Batch> {
+        self.account(list1.iter().chain(list2.iter()).map(M::account), -1);
```
frankmcsherry, line 128: Double check whether this is double accounting!

src/trace/implementations/merge_batcher.rs
```diff
@@ -173,88 +230,66 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
+        // `batch` is either a shared reference or an owned allocations.
+        let mut owned = match batch {
+            RefOrMut::Ref(vec) => {
+                let mut owned = self.empty(stash);
+                owned.clone_from(vec);
+                owned
             }
+            RefOrMut::Mut(vec) => std::mem::take(vec)
+        };
+        consolidate_updates(&mut owned);
```
frankmcsherry, line 266: Double check whether we should be collecting small batches rather than consolidating and shipping them. The concern is that once they are presented outward to the batcher it will merge batches without thinking about the number of records in them. We want to protect ourselves against quadratic behavior with lots of small elements.

src/trace/implementations/merge_batcher.rs
```diff
@@ -281,81 +317,111 @@ impl<D: Ord, T: Ord, R: Semigroup> MergeSorter<D, T, R> {
...
 
-    /// Push a batch to `self.queue` and account size changes.
-    #[inline]
-    fn queue_push(&mut self, batch: Vec<Vec<(D, T, R)>>) {
-        self.account(batch.iter().map(Vec::len), 1);
-        self.queue.push(batch);
+        for mut buffer in merged {
+            for (data, time, diff) in buffer.drain(..) {
+                if upper.less_equal(&time) {
+                    frontier.insert(time.clone());
```
frankmcsherry, line 360: Consider `insert_ref` here to avoid a clone! :D

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -214,116 +35,89 @@ impl<D: Ord+Columnation+'static, T: Ord+Columnation+'static, R: Semigroup+Column
...
             self.pending.extend(batch.drain(..std::cmp::min(batch.len(), self.pending.capacity() - self.pending.len())));
             if self.pending.len() == self.pending.capacity() {
-                crate::consolidation::consolidate_updates(&mut self.pending);
+                consolidate_updates(&mut self.pending);
                 if self.pending.len() > self.pending.capacity() / 2 {
                     // Flush if `self.pending` is more than half full after consolidation.
-                    self.flush_pending();
+                    let mut stack = self.empty(stash);
+                    stack.reserve_items(self.pending.iter());
+                    for tuple in self.pending.drain(..) {
```
frankmcsherry, line 97: Perhaps we want to copy less than everything, as this may produce irregularly shaped batches?

src/trace/implementations/merge_batcher_col.rs
```diff
@@ -214,116 +35,89 @@ impl<D: Ord+Columnation+'static, T: Ord+Columnation+'static, R: Semigroup+Column
...
-            let list2 = self.queue_pop().unwrap();
-            let merged = self.merge_by(list1, list2);
-            self.queue_push(merged);
-        }
-
-        if let Some(mut last) = self.queue_pop() {
-            std::mem::swap(&mut last, target);
-        }
+    fn finish(&mut self, _stash: &mut Vec<Self::Batch>) -> Vec<Self::Batch> {
+        vec![]
```
frankmcsherry, line 109: Don't forget to flush `self.pending`!

frankmcsherry (approved):
This generally looks good! I left some comments from our review, one of which is correctness-y, but others are nits that we can clean up as you like. On bonus ask is that perhaps we could find a name other than `Batch` to avoid clashing with pre-existing uses. We discussed `Block` or `Chunk`, neither of which are especially more insightful .. but if another name presents itself amazing! :D

PR #476, opened 2024-04-20
Remove `ValOwned`

frankmcsherry:
Thought experiment removing `ValOwned` from traces, batches, cursors, etc. It removes any opinions about the owned forms of the value GAT, other than that implied by `MyTrait::Owned` (which .. can vary as a function of the lifetime, and probably isn't load bearing at this point).

The pain here is that you do want some relationship between `V` and `Tr::Val<'a>`. All we seem to need is a function from the latter to the former, although `From` and `Into` do not work because they are not implemented for e.g. `&T` where `T: Clone` which we see a lot. Instead, core methods take a `from: Fn(Tr::Val<'_>) -> V`, which is not the most ergonomic but seemed better than having an ongoing struggle with new traits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ members = [
     # "advent_of_code_2017",
     "dogsdogsdogs",
     "experiments",
-    "interactive",
+    #"interactive",
     "server",
     "server/dataflows/degr_dist",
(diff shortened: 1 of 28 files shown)


antiguru (approved):
Looks good, Delta what we discussed!

PR #477, opened 2024-04-23
Allow joins to produce arbitrary containers

antiguru:
Add a `join_traces_core` function that allows the caller to specify the output stream container type. The existing `join_traces` function forces it to be vectors, by virtue of wrapping the output in a collection.


frankmcsherry (approved):
Looks great, but ideally we could merge `join_traces` and `join_traces_core`: the more general one is great, and the slightly less general one can probably just be skipped (as you note, it is just a `.as_collection()` call away).

PR #478, opened 2024-04-25
`join_traces` uses timely's container builder

antiguru:
Switch `join_traces` to use timely's container builder, and add a container builder that consolidates in-place.


frankmcsherry (approved):
lgtm!

PR #482, opened 2024-05-07
Seal: Drop chain while feeding data to builder

antiguru:
Fix a regression where the columnated merge batcher would only deallocate the merged chain after revealing all data to the builder.  Instead, it now drains the chain, which deallocates each chunk after revealing its contents to the builder.


frankmcsherry (approved):
Looks good, and thank you!

Issue #479, opened 2024-04-29
Suggestion: WebAssembly support


oli-w:
Hi,
I was interested in seeing if I could run differential-dataflow on the web by compiling to WebAssembly using [wasm-pack](https://github.com/rustwasm/wasm-pack) and it worked out really well! However it required some ~hacks~ changes to differential-dataflow and timely-dataflow, specifically:

* For both packages, replacing `std::time::Instant` with `web_time::Instant` - crate: [web-time](https://github.com/daxpedda/web-time) This is because there is no "current time" implemented for the `wasm32-unknown-unknown` compilation target - so we have to hook it up to the browser's [Performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) function. In all other contexts this uses the regular `std::time::Instant` under the hood so shouldn't make any difference.
* Differential dataflow has a 63 bit shift `(1 << 63)` which needs to be `(1 << 31)` in 32-bit WebAssembly.

The changes required are:
* Timely dataflow: https://github.com/TimelyDataflow/timely-dataflow/compare/master...oli-w:timely-dataflow:master
* Differential dataflow: https://github.com/TimelyDataflow/differential-dataflow/compare/master...oli-w:differential-dataflow:master

I was wondering if you would be interested in incorporating this or would prefer to keep this kind of change as a fork?

frankmcsherry:
It sounds interesting! I'll need a bit to get my head around what it entails, but it seems quite reasonable to 1. limit the use of assumptions about time, and 2. be as serious as possible about `usize` sizes.

If it's alright, I'll take a few to read up on things. Mostly, I'd love to avoid having code in TD/DD that I don't understand, because it's largely on me if something is broken elsewhere. But I have to imagine if nothing else a feature flag for time would be really easy, as should be fixing DD's 64 bit assumption (which is probably a bug anyhow).

antiguru:
I think all changes look fine from a high level. The time changes don't seem to change any behavior on non-wasm targets. The RHH code isn't used anywhere at this point. Before committing to supporting wasm, I'd like to know if there's a regression test we could have. It seems testing isn't as simple as it is for other targets, but if we can get it into CI then I don't think there's anything blocking this.

As @frankmcsherry points out, there are some parts of the code that might assume `usize` to have the same length as `u64`. Removing this assumption would be a win in itself.

frankmcsherry:
[...]

PR #485, opened 2024-05-13
Remove uses of `time::Instant` from core DD

frankmcsherry:
DD uses `time::Instant` in two non-load-bearing moments, and by removing them we open up doors for using DD in settings without support for time, including WebAssembly. This PR proposes just to remove these uses (and move one use to the commented-out code that references it).

Discussion in https://github.com/TimelyDataflow/differential-dataflow/issues/479

cc: @oli-w

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -42,47 +42,6 @@ pub struct Progress<T> {
     pub counts: Vec<(T, usize)>,
 }
 
-/// An iterator that yields with a `None` every so often.
-pub struct YieldingIter<I> {
-    /// When set, a time after which we should return `None`.
-    start: Option<std::time::Instant>,
-    after: std::time::Duration,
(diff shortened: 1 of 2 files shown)


src/capture.rs
```diff
@@ -813,6 +772,49 @@ pub mod sink {
...
+//                 None => {
+//                     self.start = None;
+//                     None
+//                 }
+//             }
+//         }
+//     }
+// }
+
+
```
antiguru, line 816: I'd prefer removing this now commented implementation, or moving it to an example together with the kafka source.

frankmcsherry: Me too! It's .. probably something that lives in `itertools` or the equivalent. Certainly not in any way specific to the code around here.

antiguru (approved):

PR #487, opened 2024-05-13
GAT time and diff

antiguru:
Turn time and diff into GATs within the storage layer, and expose the diff GAT trough cursors. This enables more efficient storage for times and diffs in arrangements, while leaving the upper layers mostly the same.


antiguru:
Closing because superseded.

PR #488, opened 2024-05-15
Use consolidating builder in consolidate_stream

antiguru:
This switches from consolidating each input individually to consolidating at the output. The benefits are that it can yield better consolidation performance because it can consolidate across input containers instead of only being able to consolidate each individual input container.


frankmcsherry (approved):
Looks great; thank you!

PR #489, opened 2024-05-16
Consolidate output diffs before producing tuples

frankmcsherry:
The `Arranged::as_collection` method converts an arrangement to a stream of its underlying updates. It does this without any thought for consolidation as it does, under the historically accurate but no longer correct take that the full arrangement detail is required. When one uses a `TraceFrontier` wrapper, which advances contained timestamps, many of these updates may now consolidate. We should do the work here, at the first moment, rather than downstream. The incremental cost means additional work for folks who don't need it, but the harm done to folks using `TraceFrontier` can be much larger.

### DIFF
--- a/src/operators/arrange/arrangement.rs
+++ b/src/operators/arrange/arrangement.rs
@@ -205,6 +205,7 @@ where
         I::Item: Data,
         L: FnMut(Tr::Key<'_>, Tr::Val<'_>) -> I+'static,
     {
+        let mut buffer = Vec::new();
         stream.unary(Pipeline, "AsCollection", move |_,_| move |input, output| {
             input.for_each(|time, data| {
                 let mut session = output.session(&time);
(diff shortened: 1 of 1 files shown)


ggevay:
In the common case where the arrangement is not a retain history index, will this consolidation still do something?

Also, will this increase the memory spikes of initial hydrations where the arrangement is not a retain history index?

antiguru (approved):
Looks good. If you're interested, we could use some of the special-casing we implemented in https://github.com/MaterializeInc/materialize/pull/27136 here too.

frankmcsherry:
> In the common case where the arrangement is not a retain history index, will this consolidation still do something?

Probably not. For an alternate approach, https://github.com/TimelyDataflow/differential-dataflow/pull/490 attaches the logic to `TraceFrontier`, so that folks who are not importing traces will not experience any overhead. 

> Also, will this increase the memory spikes of initial hydrations where the arrangement is not a retain history index?

I hope not. The buffer size is going to be proportional to the longest history of any one `(key, val)` pair, and .. we could dial that down with a `ChangeBatch` style structure, which consolidates as it goes (if the history is one that consolidates down, it would be even less consequential). Though, getting `ChangeBatch` out of timely and into differential would be a bunch of typing.

PR #491, opened 2024-05-20
Pass data from batcher to builder by chunk

antiguru:
Currently, the data shared between the batcher and the builder are individual tuples, either moved or by reference. This limits flexibility around what kind of data can be provided to a builder, i.e., it has to be in the form of tuples, either owned or a reference to a fully-formed one. This works fine for vector-like structures, but will not work for containers that like to arrange their data differently.

This change alters the contract between the batcher and the builder to provide chunks instead of individual items (it does not require _chains_.) The data in the chunks must be sorted, and subsequent calls must maintain order, too. The input containers need to implement `BuilderInput`, a type that describes how a container's items can be broken into key, value, time, and diff, where key and value can be references or owned data, as long as they can be pushed into the underlying key and value containers.

The change has some quirks around comparing keys to keys already in the builder. The types can differ, and the best solution I could come up with was to add two explicit comparison functions to `BuilderInput` to compare keys and values. While it is not elegant, it allows us to move forward with this change, without adding nightmare-inducing trait bounds all-over.

### DIFF
--- a/src/operators/reduce.rs
+++ b/src/operators/reduce.rs
@@ -5,7 +5,8 @@
 //! to the key and the list of values.
 //! The function is expected to populate a list of output values.
 
-use timely::container::{PushContainer, PushInto};
+use timely::Container;
+use timely::container::PushInto;
 use crate::hashable::Hashable;
 use crate::{Data, ExchangeData, Collection};
 use crate::difference::{Semigroup, Abelian};
@@ -313,7 +314,7 @@ where
     V: Data,
     F: Fn(T2::Val<'_>) -> V + 'static,
     T2::Batch: Batch,
-    <T2::Builder as Builder>::Input: PushContainer,
+    <T2::Builder as Builder>::Input: Container,
     ((T1::KeyOwned, V), T2::Time, T2::Diff): PushInto<<T2::Builder as Builder>::Input>,
     L: FnMut(T1::Key<'_>, &[(T1::Val<'_>, T1::Diff)], &mut Vec<(V,T2::Diff)>, &mut Vec<(V, T2::Diff)>)+'static,
 {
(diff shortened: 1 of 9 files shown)


src/operators/reduce.rs
```diff
@@ -448,10 +450,10 @@ where
                         // TODO: It would be better if all updates went into one batch, but timely dataflow prevents
                         //       this as long as it requires that there is only one capability for each message.
                         let mut buffers = Vec::<(G::Timestamp, Vec<(V, G::Timestamp, T2::Diff)>)>::new();
-                        let mut builders = Vec::new();
+                        let mut chains = Vec::new();
```
frankmcsherry, line 453: Worth noting that this is a potential regression, on account of staging all the data. You mentioned that another pattern is just to flush the buffers .. whenever one likes. I figure we should either jot that down in a comment, or make it be the case if we are worried about it.

src/trace/implementations/mod.rs
```diff
@@ -320,6 +363,109 @@ impl BatchContainer for OffsetList {
     }
 }
 
+/// Behavior to split an update into principal components.
+pub trait BuilderInput<L: Layout> {
```
frankmcsherry, line 367: We discussed whether this might work with a `BuilderInput<'a, L>`. We didn't conclude that it could or couldn't, but if it ended up being `C::Item<'a>: BuilderInput<'a, L>` that might be clearer than the alternative.

frankmcsherry: Another thought we had: Perhaps `BuilderInput<L>: Container`, and rather than introduce an `Item<'a>` here and equate it to `C::Item<'a>`, the signatures could just use `C::Item<'a>`.

frankmcsherry (approved):
I think this is good. We discussed a bit about `reduce.rs` and how to avoid potentially exotic regressions there with large keys and multiple updates. That feels like something we want to think carefully about. Other discussion point was around ergonomics for the `BuilderInput` trait, and we guessed a bit there and maybe there is a better thing to do and maybe not.

src/operators/reduce.rs
antiguru: I looked into whether it's possible to populate a builder from `((&K, V), T, R)` instead of cloning `K`. The short answer is no, the long answer is maybe. At the moment, we have two blockers:
1. Reductions write their output into an arrangement, but bypass the merge batcher. We still need to provide a merge batcher type, even tho it'll never be used. We can test removing it by providing a dummy `Batcher` trait implementation that is mostly `unimplemented!()`.
2. The input chunks to the builder need to implement `Container`, which implies `'static`. No way to store references! https://github.com/TimelyDataflow/timely-dataflow/pull/540 might help here.

I think this is something we should keep as a to-do, but not prioritize until we have evidence it causes performance issues.

frankmcsherry:
This is looking good, modulo the conflicts to resolve!

PR #492, opened 2024-05-21
Cleanup, fix rhh example

antiguru:
As the title says, cleanup and fix the RHH example. Removes the `SliceContainer2` type, which really just existed as an example on how to implement various requirements for a custom type. Good to have, but wrong place.


frankmcsherry (approved):
Amazing!

PR #495, opened 2024-05-24
Introduce and integrate `IntoOwned` trait

frankmcsherry:
This PR rebrands `MyTrait<'a>` as `IntoOwned<'a>`, an analogue of `ToOwned` that is more GAT-friendly. GATs can implement `IntoOwned<'_>` and support the re-borrowing of the owned type back into the GAT. The existing `ToOwned` infrastructure relies instead on Rust's `Borrow`, which can only borrow into a reference to a `Borrowed` type.

The PR also streamlines `MyTrait`, removing comparison functionality from it. The inequalities were unused, and the equality method can be locally implemented using `borrow_as`, as our implementors all support co-variant `PartialOrd` implementations.

Some other light changes around removing the `push` functionality for `BatchContainer` preceded this work (pointed out that the remaining complexity was `MyTrait`), but I think we agree that these changes are good too.

Only known regression is in `rhh.rs`, where previously we supported a non-allocating hash evaluation, but now call `into_owned()`. We can shake that out once things are more stable and `rhh` is actually expected to work.

### DIFF
--- a/src/trace/cursor/mod.rs
+++ b/src/trace/cursor/mod.rs
@@ -14,56 +14,39 @@ pub mod cursor_list;
 pub use self::cursor_list::CursorList;
 
 use std::borrow::Borrow;
-use std::cmp::Ordering;
 
-/// A type that may be converted into and compared with another type.
+/// A reference type corresponding to an owned type, supporting conversion in each direction.
 ///
-/// The type must also be comparable with itself, and follow the same 
-/// order as if converting instances to `T` and comparing the results.
-pub trait MyTrait<'a> : Ord {
+/// This trait is analogous to `ToOwned`, but not as prescriptive. Specifically, it avoids the
+/// requirement that the other trait implement `Borrow`, which unfortunately does not support
+/// GATs (the borrow must result in a `&Self::Borrowed` rather than a `Self::Borrowed<'_>`).
+pub trait IntoOwned<'a> {
     /// Owned type into which this type can be converted.
     type Owned;
     /// Conversion from an instance of this type to the owned type.
     fn into_owned(self) -> Self::Owned;
-    ///
+    /// Clones `self` onto an existing instance of the owned type.
     fn clone_onto(&self, other: &mut Self::Owned); 
-    /// Indicates that `self <= other`; used for sorting.
-    fn compare(&self, other: &Self::Owned) -> Ordering;
-    /// `self <= other`
-    fn less_equals(&self, other: &Self::Owned) -> bool {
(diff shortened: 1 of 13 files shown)


src/trace/cursor/mod.rs
```diff
@@ -14,42 +14,27 @@ pub mod cursor_list;
...
-    /// `self < other`
-    fn less_than(&self, other: &Self::Owned) -> bool {
-        self.compare(other) == Ordering::Less
-    }
     /// Borrows an owned instance as onesself.
     fn borrow_as(other: &'a Self::Owned) -> Self; 
 }
 
-impl<'a, T: Ord+ToOwned+?Sized> MyTrait<'a> for &'a T {
+impl<'a, T: Eq+ToOwned+?Sized> IntoOwned<'a> for &'a T {
```
frankmcsherry, line 34: `Eq` can be deleted.

src/trace/implementations/huffman_container.rs
```diff
@@ -32,12 +32,8 @@ where
...
 }
 
-impl<B> BatchContainer for HuffmanContainer<B>
-where
-    B: Ord + Clone + Sized + 'static,
-{
-    type PushItem = Vec<B>;
-    type ReadItem<'a> = Wrapped<'a, B>;
+use crate::trace::implementations::containers::Push;
+impl<B: Ord + Clone + Sized + 'static> Push<Vec<B>> for HuffmanContainer<B> {
```
antiguru, line 36: Nit: This could be `B: crate::Data + Sized` I think.

src/trace/implementations/option_container.rs
```diff
@@ -82,22 +90,22 @@ impl<'a, C: BatchContainer> Clone for OptionWrapper<'a, C> {
...
-            (Some(item1), None) => item1.equals(&Default::default()),
+            (None, Some(item2)) => item2.eq(&<C::ReadItem<'_> as IntoOwned>::borrow_as(&Default::default())),
+            (Some(item1), None) => item1.eq(&<C::ReadItem<'_> as IntoOwned>::borrow_as(&Default::default())),
             (Some(item1), Some(item2)) => item1.eq(item2)
         }
     }
 }
 impl<'a, C: BatchContainer> Eq for OptionWrapper<'a, C> where 
-C::PushItem: Default + Ord
+C::OwnedItem: Default + Ord
```
antiguru, line 105: Nit: Formatting here and below.

antiguru (approved):
Looks great, thank you!

frankmcsherry:
Thank you!

PR #497, opened 2024-05-25
Remove `BatchContainer::copy`

frankmcsherry:
The `copy` method is essentially the same as the requirement `: PushInto<Self::ReadItem<'_>>`. Trying this out to see what it looks like. Potentially it is better to have the method with a default implementation, so that calls to `copy()` clearly signal that this is the type's GAT (and potentially to remove footguns where the wrong thing is pushed in, and it doesn't end up as cheap as one might want).

The downside to the PR is that `OptionContainer` ends up with conflicting implementations, and so loses the ability to push any type that its wrapped container could push. Because that might be its own associated type. 

Anyhow, up for discussion.


antiguru (approved):
Looks good!

PR #500, opened 2024-05-26
Replace covariant PartialOrd with reborrow()

frankmcsherry:
Batch containers have required that their items are comparable across lifetimes. This seems to confound Rust when presented with the `Ord` constraint in addition. Another option is to remove the constraint, and require that the contain know how to reborrow its items, narrowing their lifetimes.

At the moment this is a thing to consider; there are other ways to achieve the same goal we think (e.g. removing `Ord` instead).

cc: @antiguru


antiguru (approved):
Looks good, thanks!

PR #501, opened 2024-05-27
Remove `'static` requirement from difference traits

frankmcsherry:
Our difference traits, `Semigroup`, `Monoid`, and `Abelian`, have been defined to require `Data`, which has been a helpful catch-all trait that includes `'static`. However, it is reasonable to want to implement these traits for lifetimed data, for example a `Cow<'a, Diff>` type.

This PR replaces `Data` with `Clone` in the requirements for the difference traits, and moves the `'static` and occasionally `Ord` requirements to those locations where they are required. This is a bit noisier, but I think a bit more "accurate".

### DIFF
--- a/src/operators/reduce.rs
+++ b/src/operators/reduce.rs
@@ -59,13 +59,13 @@ pub trait Reduce<G: Scope, K: Data, V: Data, R: Semigroup> where G::Timestamp: L
     ///          });
     /// });
     /// ```
-    fn reduce<L, V2: Data, R2: Abelian>(&self, logic: L) -> Collection<G, (K, V2), R2>
+    fn reduce<L, V2: Data, R2: Ord+Abelian+'static>(&self, logic: L) -> Collection<G, (K, V2), R2>
     where L: FnMut(&K, &[(&V, R)], &mut Vec<(V2, R2)>)+'static {
         self.reduce_named("Reduce", logic)
     }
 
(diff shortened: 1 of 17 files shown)


src/operators/reduce.rs
```diff
@@ -201,26 +201,26 @@ pub trait Count<G: Scope, K: Data, R: Semigroup> where G::Timestamp: Lattice+Ord
...
 {
-    fn count_core<R2: Abelian + From<i8>>(&self) -> Collection<G, (K, R), R2> {
+    fn count_core<R2: Ord + Abelian + From<i8> + 'static>(&self) -> Collection<G, (K, R), R2> {
         self.arrange_by_self_named("Arrange: Count")
             .count_core()
     }
 }
 
-impl<G, K: Data, T1, R: Semigroup> Count<G, K, R> for Arranged<G, T1>
+impl<G, K: Data, T1, R: Data+Semigroup+'static> Count<G, K, R> for Arranged<G, T1>
```
antiguru, line 217: Redundant static?

antiguru (approved):

src/operators/reduce.rs
frankmcsherry: Yes! Sorry, this was a weird one where because `R` shows up in the data it needed a `Data` constraint added back in.

PR #506, opened 2024-05-30
Generalize `Collection` to containers

frankmcsherry:
Starts the process of generalizing `Collection` to allow arbitrary containers in the underlying stream. At the moment this only supports those functions with no strong opinions about the container contents, and that work can start to emerge piecemeal.

### DIFF
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -10,12 +10,14 @@
 
 use std::hash::Hash;
 
+use timely::Container;
 use timely::Data;
 use timely::progress::Timestamp;
 use timely::order::Product;
 use timely::dataflow::scopes::{Child, child::Iterative};
-use timely::dataflow::{Scope, Stream};
+use timely::dataflow::Scope;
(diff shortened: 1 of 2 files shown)


src/collection.rs
```diff
@@ -37,24 +39,143 @@ use crate::hashable::Hashable;
...
+    /// # Examples
+    ///
+    /// ```
+    /// use differential_dataflow::input::Input;
+    ///
+    /// ::timely::example(|scope| {
+    ///     scope.new_collection_from(1 .. 10).1
+    ///          .map_in_place(|x| *x *= 2)
+    ///          .filter(|x| x % 2 == 1)
+    ///          .inspect_batch(|t,xs| println!("errors @ {:?}: {:?}", t, xs));
```
antiguru, line 146: Fix doc comment.

antiguru (approved):
Looks great to me, thanks!

PR #507, opened 2024-05-31
Relax constraints on `Collection` generic types

frankmcsherry:
Various relaxations that have become clearer as we start to move in container backing. This has the potential to have some fall out, as the implied constraints no longer holding mean users of DD may have to add those constraints themselves. E.g. `Semigroup` is no longer required for most collection operations, even though c'mon is it really not going to be a semigroup?


antiguru (approved):
Looks fine

PR #508, opened 2024-05-31
Remove `BatchContainer::{copy, copy_range}`.

frankmcsherry:
This PR replaces `BatchContainer::copy` with a `PushInto` requirement. We could leave the `copy` method while insisting on the `PushInto` implementation, if that provides more clarity about when we are specifically copying back a `ReadItem`, but it felt best to use the one idiom we have for implementing pushing into a container, vs having multiple implementations around (which we had).

`BatchContainer::copy_range` was removed because no one is using it. No harm adding it back, but in that case perhaps we instead want to support `PushInto` for a range type (or some wrapper around one).

### DIFF
--- a/src/trace/implementations/mod.rs
+++ b/src/trace/implementations/mod.rs
@@ -510,7 +510,7 @@ pub mod containers {
     use crate::trace::IntoOwned;
 
     /// A general-purpose container resembling `Vec<T>`.
-    pub trait BatchContainer: 'static {
+    pub trait BatchContainer: for<'a> PushInto<Self::ReadItem<'a>> + 'static {
         /// An owned instance of `Self::ReadItem<'_>`.
         type Owned;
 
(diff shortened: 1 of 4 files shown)


src/trace/implementations/mod.rs
```diff
@@ -510,7 +500,7 @@ pub mod containers {
     use crate::trace::IntoOwned;
 
     /// A general-purpose container resembling `Vec<T>`.
-    pub trait BatchContainer: 'static {
+    pub trait BatchContainer: for<'a> PushInto<Self::ReadItem<'a>> + 'static {
```
antiguru, line 503: Discussed in person: We could opt out of requiring `BatchContainer` to be `PushInto<...>` which would permit implementing containers that cannot be pushed at. For the moment, the change as implemented is probably the right thing to do, but in the future we could split the trait into `ReadBatchContainer` and `BatchContainer` where the former does not expose any functionality to modify the contents of the container. There's a similar idea in flatcontainer: https://github.com/antiguru/flatcontainer/pull/27

antiguru (approved):

frankmcsherry:
Thanks for the review!

PR #511, opened 2024-06-07
Adjust for timely TimelyDataflow/timely-dataflow#569

antiguru:
This fixes a potential bug in the consolidating container builder where it
would not clear buffers that were exposed to the user. There is no
obligation for the user to call clear, so call it when recycling.

### DIFF
--- a/src/consolidation.rs
+++ b/src/consolidation.rs
@@ -149,14 +149,40 @@ where
 {
     /// Flush `self.current` up to the biggest `multiple` of elements. Pass 1 to flush all elements.
     // TODO: Can we replace `multiple` by a bool?
+    #[cold]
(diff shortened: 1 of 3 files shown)


src/consolidation.rs
```diff
@@ -155,25 +155,25 @@ where
         let mut drain = self.current.drain(..(self.current.len()/multiple)*multiple).peekable();
         while drain.peek().is_some() {
             let mut container = self.empty.pop().unwrap_or_else(|| Vec::with_capacity(preferred_capacity));
+            container.clear();
```
antiguru, line 158: Fix a where we would retain data if the user didn't clear the container.

antiguru:
@frankmcsherry seems ok with it, so merging!

PR #512, opened 2024-06-11
Fix panic in to_vec

antiguru:
`rewind_vals` only works if there are any values.


frankmcsherry:
We should file an issue about the lack of clarity around requirements in calling these cursor methods. Certainly if the library itself cannot follow instructions, something is unclear. :D

PR #514, opened 2024-06-20
Define flat container types in terms of regions

antiguru:
Define flat container spines in terms of regions instead of concrete types. Add a `FlatKeySpineDefault` and similarly a `FlatValSpineDefault` type definition that defers to the `Containerized` trait to obtain the region of a type. This change tidies up some trait implementations where they in the past were defined in terms of concrete (owned) types instead of regions. This also allows for greater reuse of the region-based types because someone else can define a different `Containerized` trait but still use the region-based types.

I'm not too happy about the `*Default` names because it doesn't convey the meaning, but I haven't thought of a better name.

### DIFF
--- a/examples/spines.rs
+++ b/examples/spines.rs
@@ -65,9 +65,9 @@ fn main() {
                         .probe_with(&mut probe);
                 },
                 "flat" => {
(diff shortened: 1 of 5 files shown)


frankmcsherry (approved):
Looks good!

PR #518, opened 2024-07-02
Generalize flat container implementations to less specific regions

antiguru:
This change redefines some trait implementations in differential for flat containers to be in terms of regions fulfilling some requirements instead of specific regions. The current implementation makes it impossible to implement the traits for specific regions due to Rust's orphan rules, which blocks us from using custom region implementations in Materialize.

The change should be localized to traits and implementations related to flat container. It includes some overdue renaming to make traits and generic parameters slightly more understandable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ timely = {workspace = true}
 
 [workspace.dependencies]
 #timely = { version = "0.12", default-features = false }
-timely = { git = "https://github.com/TimelyDataflow/timely-dataflow", default-features = false }
+timely = { git = "https://github.com/antiguru/timely-dataflow", branch = "flatcontainer_storage", default-features = false }
 #timely = { path = "../timely-dataflow/timely/", default-features = false }
 
 [features]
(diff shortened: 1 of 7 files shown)


src/consolidation.rs
```diff
@@ -280,27 +280,29 @@ where
     }
 }
 
-impl<K, V, T, R> ConsolidateLayout for FlatStack<TupleABCRegion<TupleABRegion<K, V>, T, R>>
```
antiguru, line 283: This change is important.

src/trace/implementations/mod.rs
```diff
@@ -448,36 +448,30 @@ mod flatcontainer {
         type OffsetContainer = OffsetList;
     }
 
-    impl<K,KBC,V,VBC,T,R> BuilderInput<KBC, VBC> for FlatStack<TupleABCRegion<TupleABRegion<K,V>,T,R>>
```
antiguru, line 451: This change is important.

Cargo.toml
```diff
@@ -49,7 +49,7 @@ timely = {workspace = true}
 
 [workspace.dependencies]
 #timely = { version = "0.12", default-features = false }
-timely = { git = "https://github.com/TimelyDataflow/timely-dataflow", default-features = false }
+timely = { git = "https://github.com/antiguru/timely-dataflow", branch = "flatcontainer_storage", default-features = false }
```
antiguru, line 52: Wait for https://github.com/TimelyDataflow/timely-dataflow/pull/574 to land.

PR #519, opened 2024-09-09
Update to latest Timely

antiguru:
Bring in changes related to abomination and reference-counted addresses.


src/trace/implementations/ord_neu.rs
```diff
@@ -159,7 +159,6 @@ mod val_batch {
     ///
     /// The `L` parameter captures how the updates should be laid out, and `C` determines which
     /// merge batcher to select.
-    #[derive(Abomonation)]
```
frankmcsherry, line 162: Does this need a `#[derive(Serialize, Deserialize)]`? I don't know, and maybe we wont until we start to use it? :)

src/trace/implementations/ord_neu.rs
```diff
@@ -720,7 +719,6 @@ mod key_batch {
...
```
frankmcsherry, line 723: Same as above; removing a derived `Abomonation` without introducing `Serialize, Deserialize`. Idk the consequences.

src/trace/implementations/rhh.rs
```diff
@@ -250,8 +250,7 @@ mod val_batch {
     ///
     /// The `L` parameter captures how the updates should be laid out, and `C` determines which
     /// merge batcher to select.
-    #[derive(Abomonation)]
-    pub struct RhhValBatch<L: Layout> 
+    pub struct RhhValBatch<L: Layout>
```
frankmcsherry, line 253: Same Q as elsewhere.

frankmcsherry (approved):
All looks good to me, except a few moments where we lost an `Abomonation` derive and did not replace it with a `Serialize, Deserialize` derive. That might be fine (perhaps it was unused?), but I wanted to flag them in case they are known oopsies.

src/trace/implementations/ord_neu.rs
antiguru: We cannot easily. Serde requires the GAT to be serialize/deserialize unless we tell it otherwise.

frankmcsherry: Gotcha. And .. no particular reason to want to serialize / deserialize these, outside perhaps of interest in some DD-based persistence? I guess I forget why the `derive(Abomonation)` was there in the first place, and perhaps there was no reason anymore?

PR #527, opened 2024-10-18
Cursor based on trace's upper

### LINKED ISSUE #526: `count_total` panic with advanced trace
``` rust
    timely::execute_directly(move |worker| {
        let mut probe = ProbeHandle::new();
        let (mut input, mut trace) = worker.dataflow::<u32, _, _>(|scope| {
            let (handle, input) = scope.new_collection();
            let arrange = input.arrange_by_self();
            arrange.stream.probe_with(&mut probe);
            (handle, arrange.trace)
        });

        // ingest some batches
        for _ in 0..10 {
            input.insert(10);
            input.advance_to(input.time() + 1);
            input.flush();
            worker.step_while(|| probe.less_than(input.time()));
        }

        // advance the trace
        trace.set_logical_compaction(AntichainRef::new(&[2]));
        trace.set_physical_compaction(AntichainRef::new(&[2]));

        worker.dataflow::<u32, _, _>(|scope| {
            let arrange = trace.import(scope);
            arrange
                .count_total() // <-- panic
                .inspect(|d| println!("{:?}", d));
        });
    })
```
The above code panic:

> thread 'main' panicked at /home/tom/hdx/apps/cargo/git/checkouts/differential-dataflow-d065d23d797aa027/b5046c8/src/trace/implementations/spine_fueled.rs:150:9:
> assertion failed: PartialOrder::less_equal(&self.physical_frontier.borrow(), &upper)
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

It seems `count_total` only works for traces with default compaction frontier. Is it intentional? 
Full code is [here](https://github.com/nooberfsh/dd_count_total_panic/)

antiguru:
Fixes a bug where we'd obtain a trace's cursor based on a batch's lower frontier. The trace can be compacted past the batch's lower frontier, which violates `cursor_through`'s assumptions. Instead, we use the trace's upper to obtain the cursor.

Fixes #526.

### DIFF
--- a/src/operators/count.rs
+++ b/src/operators/count.rs
@@ -69,7 +69,8 @@ where
 
         self.stream.unary_frontier(Pipeline, "CountTotal", move |_,_| {
 
-            // tracks the upper limit of known-complete timestamps.
+            // tracks the lower and upper limits of known-complete timestamps.
+            let mut lower_limit = timely::progress::frontier::Antichain::from_elem(<G::Timestamp as timely::progress::Timestamp>::minimum());
             let mut upper_limit = timely::progress::frontier::Antichain::from_elem(<G::Timestamp as timely::progress::Timestamp>::minimum());
 
             move |input, output| {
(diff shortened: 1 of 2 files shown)


antiguru:
This is incorrect!

PR #545, opened 2024-12-06
Extract `Builder` from `Trace`

frankmcsherry:
The `Trace` trait has a `Builder` associated type, which is used automatically to build its batches. This PR extracts that type, and allows folks to use whichever builder they like, as long as it satisfies the corresponding constraints (on input and output, generally).

This further unlocks the ability to pick and choose your builder (as with your batcher), in the face of multiple container types.


antiguru (approved):
Looks great!

PR #546, opened 2024-12-06
Move `Batcher::seal` to `Builder`

frankmcsherry:
This PR moves the `seal` method from the batcher to the builder. The `seal` method was generic in builder, required a leaky abstraction of key, val, and update counts, and .. had very little to do with the batcher. This opens the door to batchers that know less about the structure of their updates, in particular about the key, val structure.


antiguru (approved):
Looks great!

PR #548, opened 2024-12-06
Remove (key, val) structure from merge batchers

frankmcsherry:
The vector and columnation merge batchers require `(key, val)` structure, but their implementation does not require it. Remove that requirement.


antiguru (approved):
:+1:

PR #550, opened 2024-12-07
Remove time from MergeBatcher

antiguru:
We can use the time declared by the Merger direction, which always needs to equal the time on MergeBatcher's Batcher implementation.


frankmcsherry (approved):
Looks great, thanks!

PR #552, opened 2024-12-08
Consolidation consolidation

frankmcsherry:
Some improvements around consolidation. 

First, the `consolidate_container` method was made into a trait method on `ConsolidateLayout`, which it required anyhow. This provides a default implementation, but allows other implementors to provide their own optimized implementation. For example, `Vec<(D, T, R)>` can simply consolidate in place and then `mem::swap`, which it does. This should make the chunkers based on `ConsolidateLayout` appropriate for `Vec`. I also renamed it `consolidate_into`, which made more sense once it took a `&mut self` argument.

Second, the implementation of `consolidate_into` was streamlined. It maintained an `Option` of the previous key and diff, but always sets this in the first iteration. We can simply take the first step and then maintain a non-optional previous key and diff. Minor, but should be simplifying.

### DIFF
--- a/src/consolidation.rs
+++ b/src/consolidation.rs
@@ -263,38 +263,30 @@ pub trait ConsolidateLayout: Container {
         permutation.extend(self.drain());
         permutation.sort_by(|a, b| Self::cmp(a, b));
 
-        // Consolidate sorted data.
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Nice!

PR #553, opened 2024-12-08
Work towards `Batcher` unification

frankmcsherry:
The `MergeBatcher` implementation is shared across multiple implementations, that re-do fairly similar work. This PR attempts to extract their differences into compact support traits they can implement, largely about how to read from and write into containers, while in the process of merging.

The PR is meant to be largely behavior preserving, with some light glitches here and there around the elegance with which one can manipulate the general objects. I'm 100% up for discussing those as part of review, and either shipping or prioritizing any fixes for them. Examples include: 1. dropping some containers in the act of merging (the final containers in a chain), 2. calling `unwrap()` far too often, rather than destructuring.

No hurry to merge, but did want to put this in place so we can start looking at it.

Although the PR reads as fairly LOC positive, several hundred are a new `columnar.rs` example meant to exercise the non-privileged access to these traits and types. Also, the PR does not yet `git rm` the columnation merger which is now dead code, nor is it yet able to get the flat container merger to build (???) so it cannot be removed.


src/trace/implementations/merge_batcher.rs
```diff
@@ -224,170 +229,466 @@ pub trait Merger: Default {
...
+                let (_data, _time, diff2) = item2;
+                stash.clone_from(diff1);
+                stash.plus_equals(&diff2);
+                if !stash.is_zero() {
+                    self.copy_destructured(data, time, stash);
+                }
+            }
+            fn account(&self) -> (usize, usize, usize, usize) {
+                let (size, capacity, allocations) = (0, 0, 0);
+                (self.len(), size, capacity, allocations)
```
antiguru, line 599: This needs to be implemented.

examples/columnar.rs
```diff
@@ -0,0 +1,453 @@
...
+            let (___d, ___t, diff2) = item2;
+            <CR as AsSpecificBytes>::clone_onto(&diff1, stash);
+            stash.plus_equals(&diff2);
+            self.push_into((data, time, &*stash));
+        }
+
+        fn account(&self) -> (usize, usize, usize, usize) {
+            let (size, capacity, allocations) = (0, 0, 0);
+            (self.len(), size, capacity, allocations)
+        }
```
antiguru, line 418: Should implement this at some point.

examples/columnar.rs
```diff
@@ -0,0 +1,453 @@
...
+        fn is_empty(&self) -> bool { self.head == self.list.len() }
+        fn cmp_heads(&self, other: &Self) -> std::cmp::Ordering {
+            let (data1, time1, _) = self.list.borrow().get(self.head);
+            let (data2, time2, _) = other.list.borrow().get(other.head);
+            (data1, time1).cmp(&(data2, time2))
+        }
+        fn from(list: Column<(CD, CT, CR)>) -> Self { Self { list, head: 0 } }
+    }
+}
```
antiguru, line 453: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files :)

src/trace/implementations/merge_batcher.rs
```diff
@@ -224,170 +229,466 @@ pub trait Merger: Default {
...
-        // TODO: Should we limit the size of `stash`?
-        if chunk.capacity() == self.chunk_capacity() {
-            chunk.clear();
+    impl<MC: MergerChunk, CQ> ContainerMerger<MC, CQ> {
+        /// Helper to get pre-sized vector from the stash.
+        #[inline]
+        fn empty(&self, stash: &mut Vec<MC>) -> MC {
+            stash.pop().unwrap_or_else(|| {
+                let mut container = MC::default();
+                container.ensure_capacity(&mut None);
```
antiguru, line 320: At the moment, we have a different capacity for columnation on the wire vs. in the merge batcher, and if I remember correctly, the reason for this is that larger columnation objects play nicer with lgalloc because their spilled information is potentially larger. I just want to point this out so we don't forget, but at the same time it's a good moment to revisit this behavior.

antiguru (approved):
I think we can land this change, but there's one comment around pre-sizing chunks. At the moment, we can use a different capacity than the one from Timely, with this change we'd fall back to Timely's `ensure_capacity`, which may or may not be what we want here. I think this can be mitigated by using a newtype pattern.

examples/columnar.rs
frankmcsherry: What I need is a link to the doc that explains why I should care as much as the rest of the internet. :D

examples/columnar.rs
frankmcsherry: Ack, though I think it is non-trivial without mapping the same four-tuple based stuff back to columnar. If I'm reading this correctly, the `Vec` implementation currently just says `(self.len(), 0, 0, 0)` so probably not a blocker for a demo `MergerChunk` implementation at the moment?

src/trace/implementations/merge_batcher.rs
frankmcsherry: We discussed, and the conclusion for now is that this is ok to merge, as one can work around it by wrapping `MC` appropriately, with something that has new opinions about what `at_capacity()` and `ensure_capacity()` should do.

src/trace/implementations/merge_batcher.rs
```diff
@@ -223,170 +228,473 @@ pub trait Merger: Default {
...
+pub use container::{VecMerger, ColMerger};
+
+pub mod container {
+
+    //! A general purpose `Merger` implementation for arbitrary containers.
+    //!
+    //! The implementation requires implementations of two traits, `ContainerQueue` and `MergerChunk`.
+    //! The `ContainerQueue` trait is meant to wrap a container and provide iterable access to it, as
+    //! well as the ability to return the container when iteration is complete.
+    //! The `MergerChukn` trait is meant to be implemented by containers, and it explains how container
```
petrosagg, line 240: small typo here
```suggestion
    //! The `MergerChunk` trait is meant to be implemented by containers, and it explains how container
```

frankmcsherry:
Performance-wise, this PR seems to maintain parity with `master`, and potentially improve slightly (though, it could be noise). On this branch,
```
Running `target/release/examples/spines 1000000 1000 old -w4`
Running ["old"] arrangement
3.868985583s    loading complete
...
7.910668916s    queries complete
8.129606833s    shut down

Running `target/release/examples/spines 1000000 1000 new -w4`
Running ["new"] arrangement
1.49757225s     loading complete
...
3.173547834s    queries complete
3.176058584s    shut down

Running `target/release/examples/columnar 1000000 1000 -w4`
1.260987s       loading complete
...
2.485150291s    queries complete
2.486176791s    shut down
```
vs on `master` (`examples/columnar` does not exist)
```
Running `target/release/examples/spines 1000000 1000 old -w4`
Running ["old"] arrangement
4.156000333s    loading complete
...
8.666894917s    queries complete
8.894178375s    shut down

Running `target/release/examples/spines 1000000 1000 new -w4`
Running ["new"] arrangement
1.5104135s      loading complete
...
3.193643833s    queries complete
3.195464916s    shut down
```
In discussion with @antiguru we've concluded that it should be ok despite a few quirks that we think we can dial in, around buffer sizing and other nits. We'll make sure that we assess the performance on MZ, and "fail forward" from this work.

PR #554, opened 2024-12-18
Update to track timely changes

frankmcsherry:
Timely changed some logging generics, in a way that was assessed as non-breaking, but was breaking. This is the fix-up, I think.


antiguru (approved):
Looks good, thanks!

PR #559, opened 2025-01-21
Define columnation chunker for all (D,T,R)

antiguru:
It was only defined for ((K,V),T,R), probably because it had some requirements on `K` and `V` in the past.


frankmcsherry (approved):
Looks good!

PR #564, opened 2025-01-28
Variable supports container streams

antiguru:
Variable supports container streams. The PR achieves this by introducing two new extension traits:
* `ResultsIn` applies `PathSummary::results_in` to every record in the stream.
* `Negate` negates the diffs in a stream.

This is in support of allowing columnar data in Materialize recursive contexts.


frankmcsherry (approved):
Looks good; thank you!

PR #570, opened 2025-02-18
Remove flatcontainer, absorb TimelyStack

antiguru:
Adjust to changes in TimelyDataflow/timely-dataflow#647 which removes
flatcontainer and columnation as Timely doesn't have any opinion around
the types and defers to downstream crates instead.

Ideally, we'd move the TimelyStack to an example, but that would be a
bigger change, so I'm holding off for the moment.


frankmcsherry:
This feels good, thank you! It seems to bring the collection of opinions on containers together, e.g. around `IntoOwned` and the `TimelyStack`, which should let us iterate on them a bit more easily.

PR #587, opened 2025-03-21
Introduce EnterTime, fix Negate

antiguru:
Introduces the EnterTime trait, which allows streams to explain to
differential how they can modify their timestamp component to match the
inner time.

Fixes the Negate trait so it can be implemented by downstream crates. Due
to Rust's rules, the local type has to come first, so the `G` parameter has
to come after the container type.

### DIFF
--- a/differential-dataflow/src/operators/enter.rs
+++ b/differential-dataflow/src/operators/enter.rs
@@ -1,0 +1,54 @@
+//! Enter a collection into a scope.
+
+use timely::Data;
+use timely::dataflow::{Scope, ScopeParent, Stream, StreamCore};
+use timely::dataflow::operators::core::{Enter, Map};
+use timely::dataflow::scopes::Child;
+use timely::progress::timestamp::Refines;
+
+/// Extension trait for streams.
+pub trait EnterTime<C, G, TInner>
+where
+    G: ScopeParent,
+    TInner: Refines<G::Timestamp>,
+{
+    /// The containers in the output stream.
+    type Container: Clone;
+
+    /// Brings a stream into a nested scope.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use timely::dataflow::Scope;
+    /// use differential_dataflow::input::Input;
+    ///
+    /// ::timely::example(|scope| {
+    ///
+    ///     let data = scope.new_collection_from(1 .. 10).1;
+    ///
+    ///     let result = scope.region(|child| {
+    ///         data.enter(child)
+    ///             .leave()
+    ///     });
+    ///
+    ///     data.assert_eq(&result);
+    /// });
+    /// ```
+    fn enter_time<'a>(&self, child: &Child<'a, G, TInner>) -> StreamCore<Child<'a, G, TInner>, Self::Container>;
+}
+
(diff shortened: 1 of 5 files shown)


differential-dataflow/src/operators/enter.rs
```diff
@@ -0,0 +1,54 @@
+//! Enter a collection into a scope.
+
+use timely::Data;
+use timely::dataflow::{Scope, ScopeParent, Stream, StreamCore};
+use timely::dataflow::operators::core::{Enter, Map};
+use timely::dataflow::scopes::Child;
+use timely::progress::timestamp::Refines;
+
+/// Extension trait for streams.
```
frankmcsherry, line 9: Nit, but doccomment could use some love.

differential-dataflow/src/operators/enter.rs
```diff
@@ -0,0 +1,54 @@
...
+pub trait EnterTime<C, G, TInner>
+where
+    G: ScopeParent,
+    TInner: Refines<G::Timestamp>,
+{
+    /// The containers in the output stream.
```
frankmcsherry, line 15: Similar nit: probably something better to say here. Perhaps
```
/// The type of container used to represent updates with `TInner` timestamps
```

differential-dataflow/src/operators/iterate.rs
```diff
@@ -169,7 +169,7 @@ where
 impl<G: Scope, D: Data, R: Abelian, C: Container + Clone + 'static> Variable<G, D, R, C>
 where
     G::Timestamp: Lattice,
-    StreamCore<G, C>: crate::operators::Negate<G, C> + ResultsIn<G, C>,
+    StreamCore<G, C>: crate::operators::Negate<C, G> + ResultsIn<G, C>,
```
frankmcsherry, line 172: I think conventionally, we usually have scope parameters first and containers later (e.g. timely's `Map` trait, as the first one I found). I'm looking at my local DD repo, and it has `Negate<G, C>` in it, where `G: Scope`. I'm not sure I understand the flipping that is going on here.

differential-dataflow/src/operators/enter.rs
```diff
@@ -0,0 +1,54 @@
+//! Enter a collection into a scope.
+
+use timely::Data;
+use timely::dataflow::{Scope, ScopeParent, Stream, StreamCore};
+use timely::dataflow::operators::core::{Enter, Map};
+use timely::dataflow::scopes::Child;
+use timely::progress::timestamp::Refines;
+
+/// Extension trait for streams.
+pub trait EnterTime<C, G, TInner>
```
frankmcsherry, line 10: As with the `Negate` comment, I think `G, TInner, C` is the more common order; certainly the scope comes first in just about all timely traits iirc (both `Map` and `Operator`, that I just checked). Timely's `Enter` trait is `Enter<G, T, C>` fwiw.

frankmcsherry (commented):
Nits on some doccomments, but a larger question about generic argument order for `EnterTime` and `Negate`, which feel like they should be flipped unless I misunderstand something.

Additional nit: I can't tell whether `EnterTime` should be redundant with timely's `Enter` trait, which does the same thing but doesn't change the container type. My sense is that `EnterUpdates` is maybe more what is intended, in that this is only meant to work for updates specifically. I'm not certain I understand whether this should be a trait on streams or a trait on containers; the stream logic seems to be mainly "bring the stream in, and change the container" and I can't think of other implementations.

PR #595, opened 2025-04-15
Compare once in minimize keys/vals

antiguru:
At the moment, we compare keys and values twice, one with less-than, and once with equals. We can instead compare once and use the result twice to avoid potentially costly comparison operations.


frankmcsherry:
I think this got handled in #596!

PR #596, opened 2025-04-17
Modernize `Cursor` API

frankmcsherry:
The `Cursor` API has historically relied on functions `foo_valid` and `foo`, so that one could independently validate a thing and without validation access it. In practice this results in two expensive calls, rather than one `get_foo` call that returns an option. Mostly, we see this when there are more complex containers that need to track down the destination of a reference, as opposed to a simple slice access.

The PR encourages more use of `get_foo` by removing their default implementations, adds `BatchContainer::get`, and .. somewhat unrelatedly updates the `CursorList` internals, following and extended #595.

Potentially we should add in default implementations of `foo` and `foo_valid`, but no need to do so at the moment, and seemingly all of our cursor implementations would like the manual implementation of these as well (there are moments where it helps to call them directly, and so it's hard to deprecate them completely).

### DIFF
--- a/differential-dataflow/src/operators/join.rs
+++ b/differential-dataflow/src/operators/join.rs
@@ -679,8 +679,7 @@ where
 
                     // populate `temp` with the results in the best way we know how.
                     thinker.think(|v1,v2,t,r1,r2| {
-                        let key = batch_key;
-                        logic(key, v1, v2, &t, r1, r2, &mut session);
+                        logic(batch_key, v1, v2, &t, r1, r2, &mut session);
                     });
 
                     // TODO: Effort isn't perfectly tracked as we might still have some data in the
--- a/differential-dataflow/src/operators/join.rs
+++ b/differential-dataflow/src/operators/join.rs
@@ -661,11 +661,11 @@ where
 
         let mut thinker = JoinThinker::new();
 
-        while batch.key_valid(batch_storage) && trace.key_valid(trace_storage) && effort < *fuel {
+        while let (Some(batch_key), Some(trace_key), true) = (batch.get_key(batch_storage), trace.get_key(trace_storage), effort < *fuel) {
 
-            match trace.key(trace_storage).cmp(&batch.key(batch_storage)) {
-                Ordering::Less => trace.seek_key(trace_storage, batch.key(batch_storage)),
-                Ordering::Greater => batch.seek_key(batch_storage, trace.key(trace_storage)),
+            match trace_key.cmp(&batch_key) {
+                Ordering::Less => trace.seek_key(trace_storage, batch_key),
+                Ordering::Greater => batch.seek_key(batch_storage, trace_key),
                 Ordering::Equal => {
 
                     use crate::IntoOwned;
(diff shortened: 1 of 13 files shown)


differential-dataflow/src/operators/join.rs
```diff
@@ -679,7 +679,7 @@ where
 
                     // populate `temp` with the results in the best way we know how.
                     thinker.think(|v1,v2,t,r1,r2| {
-                        let key = batch.key(batch_storage);
+                        let key = batch_key;
                         logic(key, v1, v2, &t, r1, r2, &mut session);
```
antiguru, line 683:
```suggestion
                        logic(batch_key, v1, v2, &t, r1, r2, &mut session);
```

differential-dataflow/src/trace/implementations/mod.rs
```diff
@@ -483,6 +483,15 @@ pub mod containers {
...
         /// Reference to the element at this position.
         fn index(&self, index: usize) -> Self::ReadItem<'_>;
+
+        /// Reference to the element at this position, if it exists.
+        fn get(&self, index: usize) -> Option<Self::ReadItem<'_>> {
+            if index < self.len() {
+                Some(self.index(index))
+            }
+            else { None }
+        }
```
antiguru, line 493: Do we want to provide a default implementation? If it all looks like this, then maybe not, otherwise we could avoid accidental performance problems by not providing it.

differential-dataflow/src/trace/implementations/mod.rs
```diff
@@ -483,6 +483,15 @@ pub mod containers {
 
         /// Reference to the element at this position.
         fn index(&self, index: usize) -> Self::ReadItem<'_>;
+
+        /// Reference to the element at this position, if it exists.
+        fn get(&self, index: usize) -> Option<Self::ReadItem<'_>> {
```
antiguru, line 488:
```suggestion
        #[inline(always)]
        fn get(&self, index: usize) -> Option<Self::ReadItem<'_>> {
```

antiguru (approved):
Looks good, thank you! You could add more inline annotations, but we can also add them later if we notice they might be needed. I wonder if we should provide a default implementation for `get` or not, and if it's not too much effort, and they're different, I'd lean towards not providing one. It's a breaking change anyway, so we don't have to worry about breaking more APIs.

differential-dataflow/src/trace/implementations/mod.rs
frankmcsherry: Mostly it provides an incremental deployment, vs breaking for everyone. Seems ok either way, but easier to merge this way.

differential-dataflow/src/operators/join.rs
frankmcsherry: Done!

frankmcsherry:
I added a `get` implementation for `Vec<T>` to avoid the default implementation, but .. at the moment at least forcing inlining and removing defaults seem out of sync with the surrounding code. Batch container implementors can put those annotations on their own implementations if they desire, but e.g. we don't do it for the `BatchContainer::{index, len}` implementations for `Vec<T>`. Happiest deferring that until some data-driven moment.

PR #614, opened 2025-06-24
Respect singleton counts in merge effort

frankmcsherry:
Our merge logic seemed to ignore the singleton optimization, and just watched the length of the `times` member, presuming that it would grow in proportion to the number of updates produced. That .. used to be true. But with the optimization `times` doesn't need to grow at all, if all updates have the same times. This is common in a snapshot, for example. This seems to have the potential to be very disruptive when such a merge happens.


antiguru (approved):
Yep, the logic seems wrong. Thanks for the fix!

PR #616, opened 2025-06-25
Use trie abstractions for batch implementations

frankmcsherry:
Our `ord_neu.rs` batch/trace implementations use a manual implementation of a short trie. Rather than have the work happen in bespoke methods we need to copy/paste, extract the logic out into trie "layers" which can be composed. For example, the "singleton optimization" for updates lived in four locations, with a fifth in `rhh.rs`. This change moves that to be one location, in the update trie layer, used by all four.

This is the first step in trying to make these types more "trie-forward", revealing their layered structure rather than living behind abstractions that conceal the structure. The goal for the moment is to get a sense for what the code looks like when you compartmentalize and modularize the logic and data. So far, pretty good!

Historically we had something similar, though it was more complicated than it needed to be. The reason seems to be that we previously had as trie layers pairs `(Vec<T>, Vec<usize>)` to indicate a list of keys and their offsets in the next layer. It turns out that `(Vec<usize>, Vec<T>)` is a better representation, with fewer cross-layer dependencies.

### DIFF
--- a/differential-dataflow/src/trace/implementations/ord_neu.rs
+++ b/differential-dataflow/src/trace/implementations/ord_neu.rs
@@ -147,11 +147,6 @@ pub mod layers {
         times: T,
         /// Concatenated ordered lists of update diffs, bracketed by offsets in `offs`.
         diffs: D,
-        /// Tracks the total number of updates contained within.
-        ///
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks great; let's go!

PR #624, opened 2025-07-14
Remove `IntoOwned`

frankmcsherry:
This PR moves the functionality of `IntoOwned` into the types that first provide access to the owned and borrowed types. The goal is to remove the freestanding requirement that a reference type should implement `IntoOwned`, as this is generally hard to introduce if it does not already exist. On the other hand, it is relatively easy for a container to provide this functionality, as it only needs to wrap existing logic (assuming it exists; if not there is a different problem). The main distinction is that we are happy to create new containers, or wrapper containers, etc. but much less happy to create new types that they contain.

There is a surprising amount of unmucking to do along the way, and this description will be updated with the narrative arcs if it is successful. For the moment, it is a draft PR revealing the work.

### DIFF
--- a/differential-dataflow/src/trace/implementations/mod.rs
+++ b/differential-dataflow/src/trace/implementations/mod.rs
@@ -98,9 +98,9 @@ pub trait Layout {
     type Target: Update + ?Sized;
     /// Container for update keys.
     // NB: The `PushInto` constraint is only required by `rhh.rs` to push default values.
-    type KeyContainer: BatchContainer + PushInto<<Self::Target as Update>::Key>;
+    type KeyContainer: BatchContainer<Owned = <Self::Target as Update>::Key> + PushInto<<Self::Target as Update>::Key>;
     /// Container for update vals.
-    type ValContainer: BatchContainer;
+    type ValContainer: BatchContainer<Owned = <Self::Target as Update>::Val>;
     /// Container for times.
     type TimeContainer: BatchContainer<Owned = <Self::Target as Update>::Time> + PushInto<<Self::Target as Update>::Time>;
     /// Container for diffs.
@@ -150,7 +150,7 @@ where
 /// Examples include types that implement `Clone` who prefer
 pub trait PreferredContainer : ToOwned {
     /// The preferred container for the type.
-    type Container: BatchContainer + PushInto<Self::Owned>;
+    type Container: BatchContainer<Owned = Self::Owned> + PushInto<Self::Owned>;
 }
 
 impl<T: Ord + Clone + 'static> PreferredContainer for T {
--- a/differential-dataflow/src/trace/implementations/mod.rs
+++ b/differential-dataflow/src/trace/implementations/mod.rs
@@ -294,6 +294,9 @@ impl BatchContainer for OffsetList {
     type Owned = usize;
     type ReadItem<'a> = usize;
 
+    fn into_owned<'a>(item: Self::ReadItem<'a>) -> Self::Owned { item }
+    fn borrow_as<'a>(owned: &'a Self::Owned) -> Self::ReadItem<'a> { *owned }
+
     fn reborrow<'b, 'a: 'b>(item: Self::ReadItem<'a>) -> Self::ReadItem<'b> { item }
 
     fn with_capacity(size: usize) -> Self {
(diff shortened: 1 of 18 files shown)


differential-dataflow/src/operators/consolidate.rs
```diff
@@ -45,20 +45,21 @@ where
     /// ```
     pub fn consolidate(&self) -> Self {
         use crate::trace::implementations::{KeyBatcher, KeyBuilder, KeySpine};
-        self.consolidate_named::<KeyBatcher<_, _, _>,KeyBuilder<_,_,_>, KeySpine<_,_,_>>("Consolidate")
+        self.consolidate_named::<KeyBatcher<_, _, _>,KeyBuilder<_,_,_>, KeySpine<_,_,_>,_>("Consolidate", |key,&()| key.clone())
     }
 
     /// As `consolidate` but with the ability to name the operator and specify the trace type.
-    pub fn consolidate_named<Ba, Bu, Tr>(&self, name: &str) -> Self
+    pub fn consolidate_named<Ba, Bu, Tr, F>(&self, name: &str, reify: F) -> Self
```
antiguru, line 52: Nit: Document `reify`.

differential-dataflow/src/trace/wrappers/frontier.rs
```diff
@@ -150,10 +153,8 @@ impl<C: Cursor> Cursor for CursorFrontier<C, C::Time> {
     fn map_times<L: FnMut(Self::TimeGat<'_>, Self::DiffGat<'_>)>(&mut self, storage: &Self::Storage, mut logic: L) {
         let since = self.since.borrow();
         let until = self.until.borrow();
-        let mut temp: C::Time = <C::Time as timely::progress::Timestamp>::minimum();
         self.cursor.map_times(storage, |time, diff| {
-            use crate::IntoOwned;
-            time.clone_onto(&mut temp);
+            let mut temp = C::owned_time(time);
```
antiguru, line 157: This might be a slight regression on account of not re-using the `temp` allocation.

dogsdogsdogs/src/operators/count.rs
antiguru, line 1: Let's leave a note somewhere that we should revisit the `IntoOwned` usage in dogs^3, but otherwise changes look OK.

differential-dataflow/src/trace/implementations/mod.rs
```diff
@@ -454,20 +499,34 @@ pub mod containers {
...
         }
+        /// Borrows an owned instance as oneself.
+        #[must_use]
+        fn borrow_as<'a>(owned: &'a Self::Owned) -> Self::ReadItem<'a>;
+
+
+        /// Push an item into this container
+        fn push_ref(&mut self, item: Self::ReadItem<'_>);
+        /// Push an item into this container
+        fn push_own(&mut self, item: Self::Owned);
```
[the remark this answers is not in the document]
antiguru: Ah, this would probably be annoying to use because we'd need to disambiguate which variant to call on each call site. I'm OK with leaving the two function as they are.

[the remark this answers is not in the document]
frankmcsherry: I think they were these and I removed them, is the change. :D It was getting in the way of creating new batch containers, in support of layouts for cursors with non-standard times (e.g. entering a loop). Though also, I think they are more clear, and it makes it easier (for me) to scan for "who is pushing owned things and why?".

So, imo more extensible (by not requiring trait bounds on types you may not control, like `(B1, B2)`, and more clear by breaking apart the two cases we expect folks to use. I prefer it, but I agree it's less "slick" than before.

differential-dataflow/src/operators/consolidate.rs
frankmcsherry: Will do.

differential-dataflow/src/trace/wrappers/frontier.rs
frankmcsherry: Done

dogsdogsdogs/src/operators/count.rs
frankmcsherry: I'm going to let this be the note! :D At least, the next step is to get rid of them, and I'm off to do that right now.

PR #627, opened 2025-07-15
WIP: `Layout` extension trait

frankmcsherry:
The intent here is to provide a `LayoutExt` extension trait for types that have a `Layout` member, to provide them with type aliases and methods on them, e.g. "owned key", "borrowed key", "borrowed to owned function".


frankmcsherry:
I'm tempted to say that the complexity (and anxiety) about the layout extension trait might justify just eating the complexity of copy/pasting its contents into each of `Cursor`, `BatchReader`, and `TraceReader`, and having the enormous impl blocks. But at least some confidence about how the trait and type resolution works. Pondering now.

PR #629, opened 2025-07-23
Inline consolidate fast-path

antiguru:
Gives the compiler the opportunity to inline consolidate calls for small
amounts of data by separating the complex consolidate logic for more than
one update to a different function. We don't mark the function as cold or
inline never because we want to leave the decision to the compiler.

In local testing, this shaved a few percentages of the spines example. With
the patch:
```
Running ["new"] arrangement
5.682871841s    loading complete
14.032609636s   queries complete
14.037398497s   shut down
```

With current master:
```
Running ["new"] arrangement
6.010566878s    loading complete
14.673966926s   queries complete
14.678923984s   shut down
```

### DIFF
--- a/differential-dataflow/src/consolidation.rs
+++ b/differential-dataflow/src/consolidation.rs
@@ -115,7 +115,7 @@ pub fn consolidate_updates_slice<D: Ord, T: Ord, R: Semigroup>(slice: &mut [(D, 
 }
 
 /// Part of `consolidate_updates_slice` that handles slices of length greater than 1.
-pub fn consolidate_updates_slice_slow<D: Ord, T: Ord, R: Semigroup>(slice: &mut [(D, T, R)]) -> usize {
+fn consolidate_updates_slice_slow<D: Ord, T: Ord, R: Semigroup>(slice: &mut [(D, T, R)]) -> usize {
     // We could do an insertion-sort like initial scan which builds up sorted, consolidated runs.
     // In a world where there are not many results, we may never even need to call in to merge sort.
     slice.sort_unstable_by(|x,y| (&x.0, &x.1).cmp(&(&y.0, &y.1)));


differential-dataflow/src/consolidation.rs
```diff
@@ -31,55 +32,61 @@ pub fn consolidate<T: Ord, R: Semigroup>(vec: &mut Vec<(T, R)>) {
...
 
 /// Sorts and consolidates a slice, returning the valid prefix length.
+#[inline]
 pub fn consolidate_slice<T: Ord, R: Semigroup>(slice: &mut [(T, R)]) -> usize {
-
     if slice.len() > 1 {
+        consolidate_slice_slow(slice)
+    }
+    else {
+        slice.iter().filter(|x| !x.1.is_zero()).count()
```
frankmcsherry, line 48: Probably not the most important thing, but I wonder if there is a way to write this that is more helpful to Rust. Like, rather than iter and hope Rust realizes the slices are either len 0 or len 1, we could `get(0)` and `unwrap_or(0)`?

frankmcsherry: Ah, I see these were inherited from the previous implementation! We can explore in follow-up work, or .. ignore if no improvements!

differential-dataflow/src/consolidation.rs
```diff
@@ -31,55 +32,61 @@ pub fn consolidate<T: Ord, R: Semigroup>(vec: &mut Vec<(T, R)>) {
...
+    }
+}
 
-        // We could do an insertion-sort like initial scan which builds up sorted, consolidated runs.
-        // In a world where there are not many results, we may never even need to call in to merge sort.
-        slice.sort_by(|x,y| x.0.cmp(&y.0));
+/// Part of `consolidate_slice` that handles slices of length greater than 1.
+fn consolidate_slice_slow<T: Ord, R: Semigroup>(slice: &mut [(T, R)]) -> usize {
```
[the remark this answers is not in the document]
frankmcsherry: Naming nit but, it's not that it's `slow` it's just general. There isn't a fast version, there's just a less general version.

frankmcsherry (approved):
Seems good! I had some nits that we can follow up with async.

PR #643, opened 2025-08-29
Adjust Differential to recent Timely changes

### LINKED ISSUE #642: Workflow "Test build against Timely master" failed
The workflow "**Test build against Timely master**" failed. See logs:
https://github.com/TimelyDataflow/differential-dataflow/actions/runs/17721617402

Commit: https://github.com/TimelyDataflow/differential-dataflow/commit/

// cc: @frankmcsherry, @antiguru — please triage and resolve this issue

antiguru:
Timely 0.24 compatibility.

Fixes #642

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,9 @@ resolver = "2"
 
 [workspace.dependencies]
 differential-dataflow = { path = "differential-dataflow", default-features = false, version = "0.16.2" }
-#timely = { version = "0.23", default-features = false }
+timely = { version = "0.24", default-features = false }
 columnar = { version = "0.10", default-features = false }
-timely = { git = "https://github.com/TimelyDataflow/timely-dataflow", default-features = false }
+#timely = { path = "../timely-dataflow/timely/", default-features = false }
 
 [profile.release]
 opt-level = 3


differential-dataflow/src/operators/reduce.rs
```diff
@@ -533,7 +533,8 @@ where
                                 for (val, time, diff) in buffers[index].1.drain(..) {
                                     buffer.push_into(((T1::owned_key(key), val), time, diff));
                                     builders[index].push(&mut buffer);
-                                    buffer.clear();
+                                    // TODO: Clear!
+                                    // buffer.clear();
```
antiguru, line 537: This is a moment where we need to think a bit:
* We could add a constraint to `Input` that we can call `clear` in it. Easy, but also not forward-thinking.
* What we're doing here is building chunks from elements. We hope that the order of items in the chunk is the same as in `buffers`, otherwise we'd construct an invalid batch. Elsewhere, we'd use a container builder that feeds into a batcher, and the batcher's output is guaranteed to be suitable for the builder.
* We're also restricting the container to absorbing _owned_ data, but we might want references here.

antiguru: Changed to use `MergerChunk::clear` here. Not ideal, but works.

differential-dataflow/src/collection.rs
```diff
@@ -155,6 +156,7 @@ impl<G: Scope, D, R, C: Container + Clone + 'static> Collection<G, D, R, C> {
     pub fn inspect_container<F>(&self, func: F) -> Self
     where
         F: FnMut(Result<(&G::Timestamp, &C), &[G::Timestamp]>)+'static,
+        C: IterContainer,
```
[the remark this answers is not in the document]
antiguru: Definitely not intended! Removed.

PR #647, opened 2025-09-21
Updates for timely 0.25

frankmcsherry:
Updates necessary for recent changes to timely, mostly around input and output handles and sessions. A bit due to removing a `+Hash` constraint on `Timestamp`.


frankmcsherry:
I think the massive addition in `examples/columnar.rs` are a merge glitch, where I developed the changes on a branch with a lot of these changes, did a `git stash`, and then popped them atop `master`. The file is almost entirely different in the branch, so .. totally mystery how this worked out. But could probably revert those changes and then start the correction again.

PR #650, opened 2025-09-24
Introduce traits for collection containers

frankmcsherry:
This PR introduces traits `Negate`, `Enter`, `Leave`, and `ResultsIn` for containers, such that collections based around containers implementing each of these traits gets access to the corresponding methods.


frankmcsherry:
In principle this should be extendable to the `Iterate` trait, but .. it's a bit weird in that the trait now needs to name two container types, the "returned" type and the "iterated" type. All possible, but a bit wordy and perhaps the next steps should be to adapt `examples/columnar.rs` to use variables first, to ensure that the traits check out.

PR #710, opened 2026-04-04
Use containers for interesting (keys, time)

frankmcsherry:
The reduce logic maintains a list `[(key, time)]` of "pending work" that cannot yet be discharged because the frontiers have yet to read the time. This has been in a `Vec<(Key, Time)>` of owned data, but is now in a container. The core logic always visited everything in order, and it is not hard to do a merge from the prior list to the next, key by key. At the moment there is still some time cloning and ownership, for ease, but this could be removed. The main goal, still in progress, is the removal of owned keys.


frankmcsherry:
So, the next challenge is that `Vec<(K, ...)>` doesn't implement `Push<(&K, ...)>` like all the better containers do. It only knows how to accept an owned `K`, or a reference to a full tuple that implements `Clone`, which makes it tricky to avoid the ownership. Probably time for the re-investigation of the `Builder` trait, and why it needs things the way it does.

PR #711, opened 2026-04-07
Allow `InternalMerger::extract` to yield when full buffer

frankmcsherry:
The `InternalMerger::extract` signature did not allow implementors to yield, and required them to drain the buffer to completion. This .. ends up problematic with some containers that need to be filled to an exact size, such as `Vec` and `TimelyStack`.

### DIFF
--- a/differential-dataflow/src/trace/implementations/merge_batcher.rs
+++ b/differential-dataflow/src/trace/implementations/merge_batcher.rs
@@ -416,19 +416,12 @@ pub mod container {
             kept: &mut Vec<Vec<(D, T, R)>>,
             stash: &mut Vec<Vec<(D, T, R)>>,
         ) {
-            use std::collections::VecDeque;
-
             let mut keep = self.empty(stash);
             let mut ready = self.empty(stash);
 
-            for chunk in merged {
-                // Convert to a VecDeque so we can pop front-by-front and
-                // yield mid-chunk when keep/ready hit capacity. Without this,
-                // pushing past `at_capacity` silently grows the buffer (Vec
-                // doubles its allocation), so the next outer at_capacity
-                // check returns false and we ship oversized chunks downstream.
-                let mut q = VecDeque::from(chunk);
-                while let Some((data, time, diff)) = q.pop_front() {
+            for mut chunk in merged {
+                // Go update-by-update to swap out full containers.
+                for (data, time, diff) in chunk.drain(..) {
                     if upper.less_equal(&time) {
                         frontier.insert_with(&time, |time| time.clone());
                         keep.push((data, time, diff));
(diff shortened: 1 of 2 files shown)


differential-dataflow/src/trace/implementations/merge_batcher.rs
```diff
@@ -402,25 +416,33 @@ pub mod container {
...
 
             for chunk in merged {
-                for (data, time, diff) in chunk {
+                // Convert to a VecDeque so we can pop front-by-front and
+                // yield mid-chunk when keep/ready hit capacity. Without this,
+                // pushing past `at_capacity` silently grows the buffer (Vec
+                // doubles its allocation), so the next outer at_capacity
+                // check returns false and we ship oversized chunks downstream.
+                let mut q = VecDeque::from(chunk);
+                while let Some((data, time, diff)) = q.pop_front() {
```
antiguru: I'm not sure I follow this part, because here we don't yield mid-chunk, don't we?

differential-dataflow/src/trace/implementations/merge_batcher.rs
```diff
@@ -638,18 +664,22 @@ pub mod container {
...
+                position: &mut usize,
                 upper: AntichainRef<T>,
                 frontier: &mut Antichain<T>,
                 keep: &mut Self,
                 ship: &mut Self,
             ) {
-                for (data, time, diff) in self.drain(..) {
+                let len = self.len();
+                while *position < len && !keep.at_capacity() && !ship.at_capacity() {
+                    let (data, time, diff) = self[*position].clone();
```
antiguru, line 673: The `clone` call is a regression over the current state.

antiguru (changes requested):
Seems the latest commit changes more than necessary?

differential-dataflow/src/trace/implementations/merge_batcher.rs
frankmcsherry: It's dead code, as it turns out. We still need the implementation for some trait bounds, but this method is not called (unless one redefines the merger for vecs).

differential-dataflow/src/trace/implementations/merge_batcher.rs
frankmcsherry: The code now tests within a chunk, is I think what the robot meant to type. I'll fix the words.

differential-dataflow/src/trace/implementations/merge_batcher.rs
frankmcsherry: Ah, I added a comment but to the wrong `.clone()`; the `merge_from` code above is full of them.

PR #714, opened 2026-04-08
Track timely `master` changes, around `Child` scope changes

frankmcsherry:
Timely introduced changes around `Child` that removes the `G: Scope` of the "parent" scope. Several attendant changes need to be shaken out, many of which are cosmetic but some of which are not. The `leave()` method becomes `leave(outer)` is the main one, and many trait bounds are changed.


frankmcsherry:
This seems all pretty good, with the call-out that there is some ergonomic grossness where
```rust
stream.scope().iterative(|inner| thing.enter(inner).leave() });
```
needs to turn in to
```rust
let outer = stream.scope();
outer.iterative(|inner| thing.enter(inter).leave(outer) });
```
This is .. not the end of the world, but is different.

antiguru (approved):
I think this looks good! Maybe we could pass `|outer, nested|` to the iterative scope, but not clear whether it works, or whether it's worth the complexity.

PR #728, opened 2026-04-24
Add batched variant of arrange for trace-less flattening

DAlperin:
Introduce a `Batched` trait that parallels `Arrange` but produces a stream of sealed batches without maintaining a shared trace. The stream is wrapped in a new `Batches<Tr>` type with inherent `as_collection` and `flat_map_ref` methods.

Split the batching half of `arrange_core` into a reusable `BatchEngine` so both `batch_core` and `arrange_core` share the capability/sealing logic. `arrange_core` also now goes through `new_trace_writer` for the trace-side half.

Add `batch_by_key` / `batch_by_self` convenience methods on `Collection` mirroring `arrange_by_*`.

Rewrite `consolidate_named` on top of `batched`: it previously built a shared trace only to immediately flatten it back into a collection, paying for a trace that nothing downstream consumed.

I spot checked a few programs compiled in release mode and the new behavior is well inlined into `arranged_core`, so it should be performance nuetral.

### DIFF
--- a/differential-dataflow/src/operators/arrange/arrangement.rs
+++ b/differential-dataflow/src/operators/arrange/arrangement.rs
@@ -171,42 +171,7 @@ impl<'scope, Tr: TraceReader> Arranged<'scope, Tr> {
             I: IntoIterator<Item: Data>,
             L: FnMut(Tr::Key<'_>, Tr::Val<'_>) -> I+'static,
     {
-        Self::flat_map_batches(self.stream, logic)
-    }
-
-    /// Extracts elements from a stream of batches as a `VecCollection`.
-    ///
-    /// The supplied logic may produce an iterator over output values, allowing either
-    /// filtering or flat mapping as part of the extraction.
-    ///
-    /// This method exists for streams of batches without the corresponding arrangement.
-    /// If you have the arrangement, its `flat_map_ref` method is equivalent to this.
-    pub fn flat_map_batches<I, L>(stream: Stream<'scope, Tr::Time, Vec<Tr::Batch>>, mut logic: L) -> VecCollection<'scope, Tr::Time, I::Item, Tr::Diff>
-    where
-        I: IntoIterator<Item: Data>,
-        L: FnMut(Tr::Key<'_>, Tr::Val<'_>) -> I+'static,
-    {
-        stream.unary(Pipeline, "AsCollection", move |_,_| move |input, output| {
-            input.for_each(|time, data| {
-                let mut session = output.session(&time);
-                for wrapper in data.iter() {
-                    let batch = &wrapper;
-                    let mut cursor = batch.cursor();
-                    while let Some(key) = cursor.get_key(batch) {
-                        while let Some(val) = cursor.get_val(batch) {
-                            for datum in logic(key, val) {
-                                cursor.map_times(batch, |time, diff| {
-                                    session.give((datum.clone(), Tr::owned_time(time), Tr::owned_diff(diff)));
-                                });
-                            }
-                            cursor.step_val(batch);
-                        }
-                        cursor.step_key(batch);
-                    }
-                }
-            });
-        })
-        .as_collection()
+        Batches { stream: self.stream }.flat_map_ref(logic)
     }
 }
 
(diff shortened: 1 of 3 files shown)


differential-dataflow/src/operators/arrange/arrangement.rs
```diff
@@ -321,157 +286,264 @@ pub trait Arrange<'scope, T: Timestamp+Lattice, C> : Sized {
...
+#[derive(Clone)]
+pub struct Batches<'scope, Tr: TraceReader> {
+    /// The stream of sealed batches.
+    pub stream: Stream<'scope, Tr::Time, Vec<Tr::Batch>>,
+}
 
-        // Acquire a logger for arrange events.
-        let logger = scope.worker().logger_for::<crate::logging::DifferentialEventBuilder>("differential/arrange").map(Into::into);
+impl<'scope, Tr: TraceReader> Batches<'scope, Tr> {
+    /// Flattens the stream of batches into a `VecCollection` with one element per `(key, val)` pair.
```
frankmcsherry, line 328: I don't think the comment is correct; there may be many `(time, diff)` pairs for each key-val pair.

differential-dataflow/src/operators/arrange/arrangement.rs
```diff
@@ -321,157 +286,264 @@ pub trait Arrange<'scope, T: Timestamp+Lattice, C> : Sized {
...
-                //    we must carve out updates now in advance of the new input frontier and
-                //    transmit them as batches, which requires appropriate *single* capabilities;
-                //    Until timely dataflow supports multiple capabilities on messages, at least.
-                //
-                // 2. If there are no held capabilities in advance of the new input frontier,
-                //    then there are no updates not in advance of the new input frontier and
-                //    we can simply create an empty input batch with the new upper frontier
-                //    and feed this to the trace agent (but not along the timely output).
-
-                // If there is at least one capability not in advance of the input frontier ...
```
frankmcsherry, line 417: All of these comments seem to have been lost. Without wanting to presume, Claude has deleted them several times for me as well, so I suspect it's not intentional.

differential-dataflow/src/operators/arrange/arrangement.rs
```diff
@@ -321,157 +286,264 @@ pub trait Arrange<'scope, T: Timestamp+Lattice, C> : Sized {
...
+///
+/// This is the batching analogue of [`Arranged`]: it carries the batch stream
+/// without the shared trace. Downstream operators can flatten the batches back
+/// into a `VecCollection` via [`Batches::as_collection`] or
+/// [`Batches::flat_map_ref`].
+#[derive(Clone)]
+pub struct Batches<'scope, Tr: TraceReader> {
+    /// The stream of sealed batches.
+    pub stream: Stream<'scope, Tr::Time, Vec<Tr::Batch>>,
+}
```
frankmcsherry, line 325: Two thoughts here:
1. TBD on whether this carries the weight of a new type; it might, or perhaps it should just be either a `type Batches = ..` or possibly just `Stream` in its uses.
2. I think it would be better if possible to have `B: BatchReader`, avoiding having to name the trace that the type is meant to avoid.

differential-dataflow/src/operators/arrange/arrangement.rs
```diff
@@ -321,157 +286,264 @@ pub trait Arrange<'scope, T: Timestamp+Lattice, C> : Sized {
...
-    // a corresponding capability, as they are only retained for actual data
-    // held by the batcher, which may prevents the operator from sending an
-    // empty batch.
-
-    let mut reader: Option<TraceAgent<Tr>> = None;
+/// This is the batching analogue of [`Arrange`]: instead of maintaining a
+/// shared trace, implementors emit sealed batches downstream as timely
+/// output. Downstream operators consume the batches directly, skipping the
+/// cost of maintaining a shared trace.
+pub trait Batched<'scope, T: Timestamp+Lattice, C> : Sized {
```
frankmcsherry, line 295: I'm not sure this trait is carrying its weight either, a bit like `Batches` may not be. The `Arranged` trait just barely makes it, because folks do want to fluently arrange things, but .. I'm not sure there are more than a handful of moments where folks want to form batches without an arrangement, and the extension traits to make the formation fluent may not need to be in the core library. Tbd!

differential-dataflow/src/operators/arrange/arrangement.rs
```diff
@@ -321,157 +286,264 @@ pub trait Arrange<'scope, T: Timestamp+Lattice, C> : Sized {
...
+    /// Produces a stream of sealed batches from `self`.
+    fn batched<Ba, Bu, Tr>(self) -> Batches<'scope, Tr>
+    where
+        Ba: Batcher<Input=C, Time=T> + 'static,
+        Bu: Builder<Time=T, Input=Ba::Output, Output = Tr::Batch>,
+        Tr: Trace<Time=T> + 'static,
```
frankmcsherry, line 301: Same as with `Bached`, if we can use `B: Batch` rather than `Tr: Trace` then probably for the best if we don't otherwise reference the trace.

frankmcsherry (commented):
A few comments left, but here's a design concern / take:

This introduces new opportunities for inconsistencies between the batches and the trace. They can now be observed "in parallel", with the arrangement former and an operator that reads the batches running not in sequence any more. The operator could receive a batch that is not yet present in the arrangement, except through good behavior of the scheduler. This is a bit unnerving, and imo would be great to avoid.

An alternate design, similar in spirit I think, is to have the `Batches` type not be a stream, but an interstitial moment that can be consumed to turn in to either a stream of batches or an `Arranged` (which also houses a stream of batches). I think it might be not unlike a shell for the batch engine, waiting to figure out what we want to do with it next.

I can imagine wanting the abstraction of "any stream of sensible batches should be able to become an arrangement", but I'm also spooked by the need to discover the new classes of concurrency that may arise in the operator implementations.

Let's discuss some more!

frankmcsherry:
Here's a concrete recommendation that I *think* makes sense: extend the `Arrange` trait with the `batched` method (perhaps `batched_named`, or just as it is) which is generic only wrt a `Ba: Batcher<Input=C>` (fingers crossed), and which produces a new type `Batched` that is .. unclear to me at the moment sorry .. but the intent is that it can either turn in to a `Stream<B::Batch>` or with an appropriate `Bu, Tr` into an `Arranged<Tr>`.

Perhaps simpler, though less fascinating: just a `batched<Ba>` method on `Arrange` that produces the stream, and then we solve for the software architecture outside of the public trait interface. I do like the idea that the parts be architected apart, but it's not clear that revealing this to users (e.g. handing a half-formed batched thing, from which they can complete the arrangement if they choose) helps them more than just saying "either batches or arrangement".

DAlperin:
Ahh I asked Claude to make me the new types, and checked the logic but not really the comments. Embarrassing!

The other points are interesting. Agree on being generic over Batch, not Tr::Batch, etc.

To give you a sense of why I landed on the design I did:
- I wanted to be able to match the fluent composition (`as_collection`/`map_batches`) of Arranged. I originally had batched variants return a stream directly but then it was subject to the StreamExt trait extensions which didn't do what I wanted here. 
- I thought it was morally nice to open the future up to the "any stream of batches could be an arrangement" thing. But to be honest I had not at all considered the concurrency issues at all. I will have to think harder about this. An interstitial object that can be exchanged for either outcomes is interesting

PR #737, opened 2026-05-04
Console based on TanStack DB

frankmcsherry:
This introduces a `console/` directory with an app that mirrors the single-page `index.html` with "more modern" components, including replication into [TanStack DB](https://github.com/tanstack/db), which has some reactive components that I am still learning about. But the intent was to remove some of the "specialized" state management, and just figure out how to replicate into a thing that supports transactional commits, from TD/DD's progress-bearing utterances.


frankmcsherry:
About 3kLOC are a json lockfile for .. lord knows what. Not as horrible as the line count might suggest. Still 1kLOC+ stuff though; not really a reduction until the console grows a bit.

PR #738, opened 2026-05-06
Remove SizableContainer requirement from InternalMerge

antiguru:
InternalMerge doesn't require SizableContainer, only some of its implementations do. Instead, we push the requirement where it's used. This allows us to build a columnar-based merge batcher without a SizableContainer implementation.


frankmcsherry:
Can merge, but it would be good to hear about any `InternalMerge` requirements; it's slated for deletion (at least, it is unused in the repo, and would only exist for demo purposes or external dependencies).

antiguru:
If it's going away, then let's just close. This was in reaction to something in another PR.

frankmcsherry:
Deletion performed in #740.

PR #749, opened 2026-06-08
interactive: import/export in the IR

frankmcsherry:
Adds `import` / `export` to the DD IR: generalizes the single `result`
output to a named export list, and adds `Node::Import` for named external
traces.

## What
- Surface: `export "name" = expr;` and `import "name";`. `result expr;`
  becomes sugar for `export "$result" = expr;`.
- IR: `Program.export: Vec<(String, Id)>` (was a single result id);
  new `Node::Import { name }`, resolved by shape against a registry.
- Threaded through `lower`, both parsers, `explain`, and the
  `ddir_vec` / `ddir_col` / `dump_explain` examples.

## Why
Foundation for multi-output programs and named-trace inputs (the upcoming
live server resolves imports against its registry); the explanation rewrite
and a future optimizer also build on the export list.

## Review notes
- `export` is fully functional standalone — the lib renderers use it.
- `import` is a *declaration with no resolver in the lib*: only a registry
  binds it, so `ddir_vec`/`ddir_col` `panic!` on a `Node::Import` and no
  example uses one yet. Compiler-IR "unresolved extern" pattern; the
  resolver lands with the server. Lib + examples only — no server crate.

### DIFF
--- a/interactive/src/explain.rs
+++ b/interactive/src/explain.rs
@@ -184,12 +184,18 @@ use clone::CloneResult;
 /// demand-set explanations for queries against the original's result.
 /// See the module doc for the architecture.
 ///
-/// `input_arities` gives `(key_arity, val_arity)` per input, in input
-/// order. Necessary because input row shapes aren't recoverable from the
-/// IR alone (Projections only invert with known input arity).
-pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
+/// `input_arities` gives `(key_arity, val_arity)` per positional input;
+/// `import_arities` gives the same per named import (entries needed for
+/// every distinct `Import { name }` referenced in `p`). Both are
+/// necessary because data-source shapes aren't recoverable from the IR
+/// alone (Projections only invert with known input arity).
+pub fn explain(
+    p: &Program,
+    input_arities: &[(usize, usize)],
+    import_arities: &BTreeMap<String, (usize, usize)>,
+) -> Program {
     let mut b = Builder::new();
-    let arities = compute_arities(p, input_arities);
+    let arities = compute_arities(p, input_arities, import_arities);
     let depths = p.depths();
     // The two user-chain lengths we track at each node:
     //
@@ -220,21 +226,37 @@ pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
     let dep_user_lens: &BTreeMap<Id, usize> = &depths;
     let n_inputs = input_arities.len();
 
+    // Distinct import names referenced by `p`, in deterministic order.
+    let import_names: Vec<String> = {
+        let mut s = std::collections::BTreeSet::new();
+        for node in p.nodes.values() {
+            if let Node::Import { name } = node { s.insert(name.clone()); }
+        }
+        s.into_iter().collect()
+    };
+
     // ---- outer scope ----
-    // Original inputs of `p`, plus one extra "query" input at index n.
+    // Original inputs of `p`, one outer-scope Import per referenced name,
+    // plus one extra "query" input.
     let original_inputs: Vec<Id> = (0..n_inputs).map(|i| b.input(i)).collect();
+    let original_imports: BTreeMap<String, Id> = import_names.iter()
+        .map(|n| (n.clone(), b.push(Node::Import { name: n.clone() })))
+        .collect();
     let query_input = b.input(n_inputs);
 
     // witness: a clone of `p`, with lift_iter chains so every witness
     // collection has a host-visible `(data, user)` form via auto-leave at
     // each enclosing user scope's exit.
-    let witness = b.clone_with_lifts(p, &original_inputs, 0);
+    let witness = b.clone_with_lifts(p, &original_inputs, &original_imports, 0);
 
     // ---- explain scope ----
     b.scope_open();
 
-    // Demand-set Variables (one per input).
+    // Demand-set Variables (one per input, one per import).
     let demand_sets: Vec<Id> = (0..n_inputs).map(|_| b.variable()).collect();
+    let import_demand_sets: BTreeMap<String, Id> = import_names.iter()
+        .map(|n| (n.clone(), b.variable()))
+        .collect();
 
     // forward inputs: demand_set_<i> | semijoin(actual_input_<i>).
     // Enter actual inputs into explain scope implicitly; semijoin restricts to
@@ -245,11 +267,18 @@ pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
             b.semijoin_data(demand_sets[i], original_inputs[i], k, v)
         })
         .collect();
+    let forward_imports: BTreeMap<String, Id> = import_names.iter()
+        .map(|n| {
+            let (k, v) = import_arities[n];
+            let semi = b.semijoin_data(import_demand_sets[n], original_imports[n], k, v);
+            (n.clone(), semi)
+        })
+        .collect();
 
     // forward: same clone procedure as witness, with substituted inputs.
     // Offset = 1 because this clone lives INSIDE the explain scope: its real
     // PointStamp depth at any point is one more than its local user_level.
-    let forward = b.clone_with_lifts(p, &forward_inputs, 1);
+    let forward = b.clone_with_lifts(p, &forward_inputs, &forward_imports, 1);
 
     // Demand Variables are pre-allocated *only* for user-program `var` IR
     // nodes (`Node::Variable`). These are the only places where the demand
@@ -271,10 +300,14 @@ pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
     // from `contribs[id]`, store it in `demand_var`, then dispatch the
     // node's bward rule (which pushes onto its inputs' contribs).
     //
-    // Query input directly seeds `contribs[result]` — the result demand
-    // starts with the query rows.
+    // Query input directly seeds `contribs[first_export]` — the demand
+    // starts with the query rows against the first export. v0 only
+    // explains a single output; multi-export programs would need one
+    // query input per export and per-output dispatch in the seeding.
+    let primary_export = p.export.first()
+        .expect("explain: program has no export to seed query input against").1;
     let mut contribs: BTreeMap<Id, Vec<Id>> = BTreeMap::new();
-    contribs.entry(p.result).or_default().push(query_input);
+    contribs.entry(primary_export).or_default().push(query_input);
 
     for (&id, node) in p.nodes.iter().rev() {
         // Scope / EndScope carry no demand and have no bward action.
@@ -319,13 +352,19 @@ pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
         b.emit_reverse(id, node, &witness, &forward, &demand_var, &arities, &host_user_lens, dep_user_lens, &mut contribs);
     }
 
-    // Bind demand-set variables: demand_set_<i> := distinct(demand_set_<i> + (demand_<Input(i)> | strip | semijoin actual)).
+    // Bind demand-set variables for inputs and imports symmetrically:
+    //   demand_set_X := distinct(demand_set_X + (demand_<X> | strip | semijoin actual)).
     // Build a Vec mapping input index `i` to its IR id in `p`, so the
-    // per-input loop below is O(n) total instead of O(n^2).
+    // per-input loop below is O(n) total instead of O(n^2). Imports are
+    // looked up by name; multiple `Import { name }` nodes in `p` share
+    // demand via the dedup pass — we route from any one of them.
     let mut input_ids: Vec<Option<Id>> = vec![None; n_inputs];
+    let mut import_ids: BTreeMap<String, Id> = BTreeMap::new();
     for (&id, node) in &p.nodes {
-        if let Node::Input(i) = node {
-            input_ids[*i] = Some(id);
+        match node {
+            Node::Input(i) => { input_ids[*i] = Some(id); }
+            Node::Import { name } => { import_ids.entry(name.clone()).or_insert(id); }
+            _ => {}
         }
     }
     for i in 0..n_inputs {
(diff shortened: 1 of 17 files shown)


interactive/src/parse/applicative.rs
```diff
@@ -40,6 +40,26 @@ fn tokenize(input: &str) -> Vec<Token> {
             '+' => { chars.next(); tokens.push(Token::Plus); },
             '*' => { chars.next(); tokens.push(Token::Star); },
             '$' => { chars.next(); tokens.push(Token::Dollar); },
+            '"' => {
+                chars.next();
+                let mut s = String::new();
+                loop {
```
frankmcsherry: As discussed, there is lexer creep here that is unused. Ideally trim so that it needn't be reviewed, but also that time has now been burned independently. Independently, avoid `loop` and other unstructured control-flow idioms that conceal the control flow structure. At the very least, indicate under which conditions the loop breaks, and satisfying which properties (what would be the `while` condition?).

interactive/src/parse/mod.rs
```diff
@@ -65,5 +68,9 @@ pub enum Stmt {
     Let(String, Expr),
     Var(String, Expr),
     Scope(String, Vec<Stmt>),
+    /// `result expr;` — sugar for `Export("$result", expr)`. Kept so
+    /// pre-export programs continue to parse.
```
frankmcsherry: There are no stability guarantees or external programs we need to prevent from breaking. If this is now vestigial best to remove it to prevent others from using it, and us from continuing to support it.

interactive/src/parse/pipe.rs
```diff
@@ -39,6 +39,26 @@ fn tokenize(input: &str) -> Vec<Token> {
             '<' => { chars.next(); if chars.peek() == Some(&'=') { chars.next(); tokens.push(Token::LtEq); } else { tokens.push(Token::Lt); } },
             '>' => { chars.next(); if chars.peek() == Some(&'=') { chars.next(); tokens.push(Token::GtEq); } else { tokens.push(Token::Gt); } },
             '$' => { chars.next(); tokens.push(Token::Dollar); },
+            '"' => {
+                chars.next();
+                let mut s = String::new();
+                loop {
```
frankmcsherry: Likely additional lexer creep, as in the applicative parser, and the same comment about `loop` concealing control flow. Change to `while` or better document what is going on here.

interactive/src/explain.rs
```diff
@@ -184,12 +184,18 @@ use clone::CloneResult;
...
-/// order. Necessary because input row shapes aren't recoverable from the
-/// IR alone (Projections only invert with known input arity).
-pub fn explain(p: &Program, input_arities: &[(usize, usize)]) -> Program {
+/// `input_arities` gives `(key_arity, val_arity)` per positional input;
+/// `import_arities` gives the same per named import (entries needed for
+/// every distinct `Import { name }` referenced in `p`). Both are
+/// necessary because data-source shapes aren't recoverable from the IR
+/// alone (Projections only invert with known input arity).
+pub fn explain(
+    p: &Program,
```
frankmcsherry, line 193: Let's find a better name than `p`, which is unclear elsewhere. Consider `prog` or `program`.

interactive/src/explain.rs
```diff
@@ -426,13 +478,17 @@ mod builder {
...
-mod arities {
+pub mod arities {
     use std::collections::BTreeMap;
 
     use crate::ir::{Id, LinearOp, Node, Program};
     use crate::parse::Reducer;
 
-    pub(super) fn compute_arities(p: &Program, input_arities: &[(usize, usize)]) -> BTreeMap<Id, (usize, usize)> {
+    pub fn compute_arities(
+        p: &Program,
```
frankmcsherry, line 488: Same as above; choose a more legible name; `prog` or `program`, for example.

interactive/src/explain.rs
```diff
@@ -512,6 +572,7 @@ mod clone {
             &mut self,
             p: &Program,
```
frankmcsherry, line 573: I see now that illegibility is precedent in this file. In that case let's leave things as `p: &Program` for consistency, with a future goal of improving the legibility (I'm sure it's generally a dog's breakfast in here).

interactive/src/ir.rs
```diff
@@ -56,6 +56,9 @@ pub enum LinearOp {
 /// Symbolic IR node.
 pub enum Node {
     Input(usize),
+    /// A named external trace, resolved against a registry at install time.
+    /// Shape is inferred from the registry, not from the IR.
+    Import { name: String },
```
frankmcsherry: My understanding is that this is not yet real, but a stub for future work. You argued for its inclusion, which is fine, but we do need to at least be clear that 1. it is not real, and 2. there is a plan which is to provide a uniform substrate that at least replaces `Input(usize)`.

interactive/src/ir.rs
```diff
@@ -72,7 +75,10 @@ pub enum Node {
 
 pub struct Program {
     pub nodes: BTreeMap<Id, Node>,
-    pub result: Id,
+    /// Named outputs of the program. A renderer (or the server's registry)
+    /// looks up each by name. The conventional single-result name is
+    /// `"$result"`, which the surface syntax `result expr;` lowers to.
+    pub export: Vec<(String, Id)>,
```
frankmcsherry: The first line, "Name outputs of the program" makes sense. The rest does not, and should be deleted. In particular, if `result` is now redundant with `export` we needn't discuss this relationship.

interactive/src/ir.rs
```diff
@@ -141,7 +150,8 @@ impl Program {
                 Node::Concat(ids) => ids.clone(),
                 Node::Leave(id, _) => vec![*id],
                 Node::Bind { value, .. } => vec![*value],
-                Node::Input(_) | Node::Variable | Node::Scope | Node::EndScope => vec![],
+                Node::Input(_) | Node::Import { .. } | Node::Variable
+                | Node::Scope | Node::EndScope => vec![],
```
frankmcsherry: Leave as one line, thank you.

interactive/src/ir.rs
```diff
@@ -195,10 +205,13 @@ impl Program {
                     if *variable == from { *variable = to; }
                     if *value == from { *value = to; }
                 },
-                Node::Input(_) | Node::Variable | Node::Scope | Node::EndScope => {},
+                Node::Input(_) | Node::Import { .. } | Node::Variable
+                | Node::Scope | Node::EndScope => {},
```
frankmcsherry: Same as above.

interactive/src/ir.rs
```diff
@@ -195,10 +205,13 @@ impl Program {
...
             }
         }
-        if self.result == from { self.result = to; }
+        for (_, id) in self.export.iter_mut() {
+            if *id == from { *id = to; }
+        }
```
frankmcsherry, line 218: Explain what is going on here, either with comments or better names.

interactive/src/explain.rs
```diff
@@ -758,7 +827,9 @@ mod reverse {
             let side = |inp: Id| Side::for_input(inp, witness, forward, arities, host_user_lens, dep_user_lens);
 
             match node {
-                Node::Input(_) => { /* terminal; feeds demand-set seeding. */ }
+                Node::Input(_) | Node::Import { .. } => {
+                    /* terminal; feeds demand-set seeding. */
```
frankmcsherry: Keep as one line.

interactive/src/lib.rs
```diff
@@ -3,33 +3,42 @@ pub mod ir;
...
+
 use parse::{Stmt, Expr};
 
-/// Count the number of distinct inputs referenced in a program.
-pub fn count_inputs(stmts: &[Stmt]) -> usize {
-    let mut max_input = 0usize;
+/// Survey the references in a program: the count of positional inputs
+/// (one more than the largest `input N` index, zero if none appear) and
+/// the set of names referenced by `import "name"`.
+pub fn count_inputs(stmts: &[Stmt]) -> (usize, BTreeSet<String>) {
```
frankmcsherry: It seems like this is now more than "count inputs"; let's improve the name to match its behavior. It seems like this is related to the confusion introduced by `import` not yet replacing `input`. If that's right, we can explain this and leave a hook for future revision when we cut over.

interactive/src/lib.rs
```diff
@@ -3,33 +3,42 @@ pub mod ir;
...
+fn walk_stmts(stmts: &[Stmt], positional: &mut usize, imports: &mut BTreeSet<String>) {
     for stmt in stmts {
         match stmt {
-            Stmt::Let(_, expr) | Stmt::Var(_, expr) | Stmt::Result(expr) => {
-                max_input = max_input.max(count_inputs_expr(expr));
-            },
-            Stmt::Scope(_, body) => {
-                max_input = max_input.max(count_inputs(body));
+            Stmt::Let(_, expr) | Stmt::Var(_, expr) | Stmt::Result(expr)
+                | Stmt::Export(_, expr) => {
```
frankmcsherry: These line breaks are not helpful. Single-line please.

interactive/src/lower.rs
```diff
@@ -51,20 +51,25 @@ impl Lowering {
...
         // ---- 1. Bucket statements; reject duplicate names. ----
         // `order` records the original textual order so the topological pass
         // is deterministic when several items are simultaneously ready.
         let mut vars: Vec<(String, Expr)> = Vec::new();
         let mut lets: HashMap<String, Expr> = HashMap::new();
         let mut scopes: HashMap<String, Vec<Stmt>> = HashMap::new();
         let mut order: Vec<(ItemKind, String)> = Vec::new();
-        let mut results: Vec<Expr> = Vec::new();
+        // Exports in their declaration order; `result expr;` is sugar for
+        // `export "$result" = expr;`.
```
frankmcsherry: Repeatedly announcing the sugar is more evidence we should just remove `result` if possible.

interactive/src/lower.rs
```diff
@@ -82,10 +87,18 @@ impl Lowering {
...
-                Stmt::Result(expr) => results.push(expr),
+                Stmt::Result(expr) => local_exports.push(("$result".into(), expr)),
+                Stmt::Export(name, expr) => local_exports.push((name, expr)),
+            }
+        }
+        {
+            let mut names: BTreeSet<&str> = BTreeSet::new();
+            for (n, _) in &local_exports {
+                if !names.insert(n) {
+                    panic!("Duplicate export name at the same scope level: {:?}", n);
```
frankmcsherry: This seems like a check for an undocumented (afaict) invariant, and is some amount of confusing because I wouldn't imagine you would want to support exporting repeated names that happen to be at different scope levels (exporting anything other than a trace at the root scope seems problematic).

frankmcsherry (commented):
Comments for Claude about things to tighten.

interactive/src/explain.rs
frankmcsherry: Left as `p` for consistency with the rest of the file, per your follow-up; noted as a future legibility pass.

interactive/src/explain.rs
frankmcsherry: Left as `p` for consistency with the rest of the file, per your follow-up; noted as a future legibility pass.

interactive/src/explain.rs
frankmcsherry: Agreed — leaving `p` for consistency, with a future legibility pass over this file.

interactive/src/parse/applicative.rs
frankmcsherry: Trimmed: dropped the unused escape handling, and replaced the bare `loop` with `while let` + a `closed` flag so the exit is explicit (read to the closing quote; error on EOF). e5d28c3f

interactive/src/parse/mod.rs
frankmcsherry: Gone: `result` is removed entirely, so `export` is now the only output statement. e5d28c3f

interactive/src/parse/pipe.rs
frankmcsherry: Same as the applicative lexer — escapes removed, `while let` + `closed` flag for an explicit exit. e5d28c3f

interactive/src/ir.rs
frankmcsherry: Doc'd as a STUB: only the server resolves it, no example uses it, with the plan to become one named-source substrate that subsumes `Input(usize)` — no two ways to bring in a source. e5d28c3f

interactive/src/ir.rs
frankmcsherry: Trimmed to just 'Named outputs of the program.' — the `result` relationship is moot now that `result` is removed. e5d28c3f

interactive/src/ir.rs
frankmcsherry: Back to one line. e5d28c3f

interactive/src/ir.rs
frankmcsherry: Back to one line. e5d28c3f

interactive/src/ir.rs
frankmcsherry: `rewrite` doc enriched: redirects every reference `from`->`to` across nodes' inputs and the export list, used by `optimize` when it collapses/fuses a node. e5d28c3f

interactive/src/explain.rs
frankmcsherry: One line now. e5d28c3f

interactive/src/lib.rs
frankmcsherry: Renamed to `survey_sources`; doc notes it returns both positional inputs and imports until `import` subsumes `input` (the cutover hook). e5d28c3f

interactive/src/lib.rs
frankmcsherry: Single-lined (and the `result` arm is gone with `result`). e5d28c3f

interactive/src/lower.rs
frankmcsherry: Removed along with `result`. e5d28c3f

interactive/src/lower.rs
frankmcsherry: Exports are now root-only — a nested `export` panics (negative-tested); the per-scope dedup is now a plain 'Duplicate export name' at the root. e5d28c3f

PR #751, opened 2026-06-08
interactive: scope/region builders

frankmcsherry:
Adds scope structure + a lifetime to builder to avoid timestamp index errors.
Specifically, the manual-depth-arithmetic bug class behind the level errors in the previous PR.

## What
- `Program::regions()` — explicit per-node) from
  the `Scope`/`EndScope` markers, plus an iterative-vs-structural tag per
  region. Structured companion to
- `region_builder` — lifetime-bounded builder whose generative-invariant brand
  makes a scope leak or a missing proven by
  `compile_fail` doctests); tracks `Shape{k,v}`; ports `reach` as a worked
  example.
- Region-IR kickoff doc.

## Why
The two level errors were manual dn. This makes
scope discipline structural (the type enforces well-formedness) and is the
foundation for a scope-oriented IR

## Notes
- Additive — new modules + one `lib.rs` line; no existing behavior changes.

### DIFF
--- a/interactive/src/ir.rs
+++ b/interactive/src/ir.rs
@@ -85,6 +85,31 @@ pub struct Program {
     pub export: Vec<(String, Id)>,
 }
 
+/// Identifier for a scope (a scope nesting). Scope 0 is the root.
+pub type ScopeId = usize;
+
+/// Structural view of the program's scope nesting, derived from the
+/// `Scope`/`EndScope` markers — the explicit form of what `depths()` only
+/// counts. Scope 0 is the top level; each `Scope` opens a child scope.
+///
+/// `iterative[r]` is true when scope `r` carries an iteration feedback (a
+/// `Variable`); a scope with none is purely *structural* — a nesting
+/// boundary that does not advance the timestamp. The scope-typed builder
+/// (see `scope_builder`) leverages the latter for cheap, lifetime-
+/// based scoping discipline, and a viz can draw scopes as nested boxes.
+#[derive(Debug, Clone)]
+pub struct ScopeTree {
+    /// Parent of each scope; `None` for the root.
+    pub parent: Vec<Option<ScopeId>>,
+    /// Direct children of each scope, in the order they were opened.
+    pub children: Vec<Vec<ScopeId>>,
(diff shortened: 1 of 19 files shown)


interactive/src/ir.rs
```diff
@@ -85,6 +85,31 @@ pub struct Program {
...
+/// Identifier for a region (a scope nesting). Region 0 is the root.
+pub type RegionId = usize;
+
+/// Structural view of the program's scope nesting, derived from the
+/// `Scope`/`EndScope` markers — the explicit form of what `depths()` only
+/// counts. Region 0 is the top level; each `Scope` opens a child region.
+///
+/// `iterative[r]` is true when region `r` carries an iteration feedback (a
+/// `Variable`); a region with none is purely *structural* — a nesting
+/// boundary that does not advance the timestamp. The region-typed builder
```
frankmcsherry: This sounds error-prone, in that the presence/absence of a variable determines timestamp offsetting, it sounds like. Removal of a variable could invalidate timestamp offset math. If that's correct, I think we should aim for something more explicit, as in TD/DD where `region` explicitly doesn't adjust the timestamp. Similarly, I think maybe `Regions` is a poor name for scopes that are not the same as regions in TD/DD (they will be implemented with regions, because the timestamp does not change in our implementation, but .. is "scopes" a bad name for some reason?).

frankmcsherry (commented):
Notes for Claude.

interactive/src/ir.rs
frankmcsherry: Agreed — renamed throughout: `RegionId`->`ScopeId`, `Regions`->`ScopeTree`, `regions()`->`scopes()`, the `region_builder` module/file->`scope_builder`, and the brand token `Region`->`Scope`. ('region' would also collide with timely's narrower meaning, so 'scope' is the consistent umbrella.) Done in f54b9584; #750 also picked up two comment refs that pointed at the builder.

frankmcsherry:
Closing as not quite ready yet. Good idea, but it isn't wired up beyond tests. Will return with a new PR once this has been wired up.

PR #765, opened 2026-06-20
Fix off-by-one in dogsdogsdogs lookup_map (index(1) -> index(0))

oflatt:
`dogsdogsdogs::operators::lookup_map` stages the probe key into a capacity-1 `KeyContainer`, pushes one element (valid index 0), but reads it back at `index(1)` — out of bounds — for both the `seek_key` and `get_key` comparisons. This panics on *every* probe, so any `lookup_map`-based operator (e.g. the worst-case-optimal delta-query path `extend` → `propose`/`count`/`validate` → `lookup_map`) is unusable.

**Fix:** read the staged key at `index(0)`.

**Test:** adds `dogsdogsdogs/tests/lookup_map_regression.rs`, which drives a triangle worst-case-optimal join over a small in-memory triangle — it panics before the fix and finds `(0, 1, 2)` after.

Found while building a worst-case-optimal join backend on differential dataflow for an e-graph / Datalog engine.

### DIFF
--- a/dogsdogsdogs/src/operators/lookup_map.rs
+++ b/dogsdogsdogs/src/operators/lookup_map.rs
@@ -94,8 +94,8 @@ where
                         if !frontier2.less_equal(time) {
                             logic2(prefix, &mut key1);
                             key_con.clear(); key_con.push_own(&key1);
-                            cursor.seek_key(&storage, key_con.index(1));
-                            if cursor.get_key(&storage) == Some(key_con.index(1)) {
+                            cursor.seek_key(&storage, key_con.index(0));
+                            if cursor.get_key(&storage) == Some(key_con.index(0)) {
                                 while let Some(value) = cursor.get_val(&storage) {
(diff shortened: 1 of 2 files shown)


oflatt:
For context, I'm working on an experimental version of egglog that uses flowlog + dogsdogsdogs as the backing database engine

antiguru:
I can confirm that I found exactly the same problem, and my agent fixed it the same way. Only that I didn't put up a PR :) Thank you!

frankmcsherry:
Yup; looks very plausible. I'm up for merging, but I'm going to need to rethink the testing posture. Much of this is not actively tested / maintained (vs the `half_join` cohort, which is). Seems like a good reason to get ducks back in a row. Is it right that the test is mostly just the example, but driven with actual data?

oflatt:
Yeah the test is a small example- I ran into a bug with a much larger egglog test

PR #790, opened 2026-07-08
join: reshape JoinTactic to an iterator of output containers

frankmcsherry:
Replaces the two-method `JoinTactic` (`defer` + fuel-metered `work`) with a single

```rust
fn prep(&mut self, input0: Vec<B0>, input1: Vec<B1>, fresh: Fresh, meet: B0::Time)
    -> Box<dyn Iterator<Item = CB::Container>>;
```

The tactic now maps two batch lists to an iterator of output containers and holds neither capabilities nor a fuel budget — the same data-to-data division of labor as `ReduceTactic`. The iterator is the suspension mechanism: a container the driver is free to stop reading.

### What moves where

- **Driver** owns the two per-direction queues, pairs each iterator with its shipping capability, pulls under a split fuel budget, ships each yielded container, and drops a unit when its iterator goes dry.
- **"Work remains"** becomes driver-observable (an iterator has yet to yield `None`) instead of a fuel-sign protocol the tactic must self-report. The footgun documented on the old `work` method is gone.
- **Fairness** (the two-queue split by arrival direction) moves to the driver.

### Sink change

Join closures now receive `&mut CB` and call `push_into` directly, rather than a timely `JoinSession`. A `Session` is bound to a live operator output and is not locally constructible, so it can't back a self-contained iterator. The arbitrary-container capability #478 wanted comes from the `ContainerBuilder` itself, not the session, so handing the closure the bare builder suffices. `JoinSession` is removed.

The `EffortBuilder` wrapper is also removed: its record counter existed only to meter fuel from a closure's output volume, and the driver now meters fuel from shipped containers' `record_count()`, leaving the type nothing to do but rename `push_into` to `give`.

### Cursor tactic

`CursorTactic` shares its logic closure across units via `Rc<RefCell<_>>` — each unit is a `'static` iterator and cannot borrow the tactic — preserving the single-mutable-state semantics of one closure threaded through every match. `Deferred` becomes `DeferredIter: Iterator`; `next()` plays the same merge-join loop forward (still `#[inline(never)]`), suspending at container boundaries rather than under a fuel budget.

### Tests

Whole workspace builds warning-free. All `differential-dataflow` tests pass, including SCC and BFS — iterative, recursive-scope dataflows, so capabilities are released correctly when iterators drain (no deadlock).

### Open for review

- **`DeferredIter::next` uses `mem::take` on `extract`** (needs `Container: Default`), stealing the builder's recycled allocation; a stash-swap would avoid the churn.
- **`logic`'s public sink type changed** (`JoinSession` → `&mut CB`); Mz exposure unverified.

### DIFF
--- a/differential-dataflow/examples/columnar/main.rs
+++ b/differential-dataflow/examples/columnar/main.rs
@@ -90,6 +90,7 @@ fn main() {
 /// Push on UpdatesTyped for the reduce builder path.
 mod reachability {
 
+    use timely::container::PushInto;
     use timely::order::Product;
     use differential_dataflow::Collection;
     use differential_dataflow::AsCollection;
@@ -149,7 +150,7 @@ mod reachability {
                         use differential_dataflow::difference::Multiply;
                         let dst: Node = *dst;
                         let diff: Diff = d1.clone().multiply(d2);
-                        session.give::<(Node, (), IterTime, Diff)>((dst, (), time.clone(), diff));
+                        session.push_into((dst, (), time.clone(), diff));
                     },
                 ).as_collection();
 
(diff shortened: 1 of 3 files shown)


frankmcsherry:
> DeferredIter::next uses mem::take on extract (needs Container: Default), stealing the builder's recycled allocation; a stash-swap would avoid the churn.

This is fixable, but seemingly pre-Polonius requires a bit of a dance to return the `&mut` borrow. A grim `loop { .. }` and a `done` variable to bail out of it.

PR #807, opened 2026-07-20
Arc-backed spines for cross-thread arrangement sharing

antiguru:
## Summary

Adds the ability to read an arrangement from a thread other than the one that maintains it, in additive commits.

### `trace: add Arc-backed spine variants for cross-thread sharing`

Adds `arc_blanket_impls` (a mirror of `rc_blanket_impls` using `Arc`) and the `ArcOrdValSpine`/`ArcOrdKeySpine` spines with their `ArcOrdValBuilder`/`ArcOrdKeyBuilder`, re-exported as `ArcValSpine`/`ArcKeySpine`/`ArcValBuilder`/`ArcKeyBuilder`.

This is **purely additive**. The default `OrdValSpine`/`OrdKeySpine`, the `RcOrd*Builder`s, and the `Val*`/`Key*` aliases stay `Rc`-backed, so existing users are unaffected. An `Arc`'d batch whose contents are `Send + Sync` can be handed to and read from another thread, which the `Rc` variants cannot. Atomic reference counting is marginally more expensive (per batch clone/drop, never per record), so `Rc` remains the default and callers opt into `Arc` only when they need the sharing.

### `arrange: share arrangements across timely runtimes`

Adds `operators::arrange::sharing`, a primitive for reading an arrangement from outside the owning worker, built on the `Arc`-backed spines:

- `Arranged::publish` attaches a publisher operator to an arrangement and returns a `Published`, whose `handle()` mints `Clone + Send` `SharedTraceHandle`s.
- `SharedTraceHandle` implements `TraceReader`, so downstream operators drive its compaction and acquire cursors as with any trace handle.
- `SharedTraceHandle::snapshot_at` serves a consistent point/full-scan read from any thread, blocking until the publication frontier passes the query time and gating on `since <= t` so a read past compaction fails rather than returning stale rows.
- `SharedTraceHandle::import` replays the shared arrangement into another scope, so a second timely runtime can consume it as an ordinary `Arranged`.

The publisher forwards the **meet** of reader holds to its own `TraceAgent` (the sole writer of the trace's compaction), never the empty meet of zero holds. Handle clones carry independent holds. Compaction is applied to the agent outside the shared lock, since physical compaction can run an unbounded merge. The published `since` is the frontier the trace will actually hold after forwarding, so a concurrently registering handle cannot latch a floor below the trace's real compaction. A late importer registering after the publisher closed seeds its own terminal frontier rather than hanging.

### `arrange/sharing: import asserts equal total peers`

Records the publisher's total peer count at publish time and asserts an importing scope's `peers()` matches it. Pairwise import (importer worker `i` reads publisher worker `i`) is sound only when both sides shard by `key.hashed() % peers`, which requires equal *total* peers (workers-per-process times processes), not just equal workers-per-process. A mismatch now panics at import time instead of silently reading the wrong shard.

### `arrange/sharing: publishing no longer pins compaction`

Previously the publisher held its own `TraceAgent` at the publish-time `since`, which never advanced, so merely publishing an arrangement froze the trace's compaction for the life of the arrangement even with zero readers. The publisher now sources the writer-driven compaction frontier from the trace box (the meet of all *other* handles' holds, i.e. the box's counted meet minus the publisher's own contribution) and forwards that when no reader holds exist, and the meet of reader holds when they do. Publishing carries no independent compaction floor: only a live importer's own `as_of` hold holds the trace back, released on drop. The published `since` equals the trace's real compaction, so a live reader can still read at times it holds.

## Motivation

This unblocks sharing an arrangement between two adjacent timely runtimes in one process (for example, a maintenance runtime and a latency-sensitive query runtime), and reading arrangement snapshots from non-timely threads.

## Testing

- `tests/trace.rs`: reads an `Arc`'d batch from another thread.
- `tests/sharing.rs`: a cross-thread snapshot, a cross-dataflow import through a shared handle, `publish_without_readers_does_not_pin_compaction`, `import_hold_pins_then_releases`, and `import_asserts_equal_peers`.
- Full existing test suite passes; the `Rc` default path is unchanged.

### DIFF
--- a/differential-dataflow/src/operators/arrange/agent.rs
+++ b/differential-dataflow/src/operators/arrange/agent.rs
@@ -568,6 +568,15 @@ pub mod trace_box {
         ///
         /// This is used to inspect batches for purposes of resource accounting in external systems.
         pub fn trace(&self) -> &Tr { &self.trace }
+        /// The accumulated logical-compaction holds of all referees of the trace.
+        ///
+        /// This is the `MutableAntichain` whose frontier the trace compacts to. Exposed so an
+        /// external referee (for example a cross-runtime publisher) can compute the frontier the
+        /// trace would compact to without its own hold, by cloning this and subtracting that hold.
+        pub fn logical_compaction(&self) -> &MutableAntichain<Tr::Time> { &self.logical_compaction }
+        /// The accumulated physical-compaction holds of all referees of the trace. See
+        /// [`TraceBox::logical_compaction`].
+        pub fn physical_compaction(&self) -> &MutableAntichain<Tr::Time> { &self.physical_compaction }
         /// Replaces elements of `lower` with those of `upper`.
         #[inline]
         pub fn adjust_logical_compaction(&mut self, lower: AntichainRef<Tr::Time>, upper: AntichainRef<Tr::Time>) {
--- a/differential-dataflow/src/operators/arrange/mod.rs
+++ b/differential-dataflow/src/operators/arrange/mod.rs
@@ -63,6 +63,7 @@ type TraceAgentQueueWriter<Tr> = Weak<(Activator, RefCell<BatchQueue<Tr>>)>;
 pub mod writer;
 pub mod agent;
 pub mod arrangement;
+pub mod sharing;
 
 pub mod upsert;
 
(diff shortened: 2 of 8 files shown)


antiguru:
Turns out we don't need this and can accomplish the same using a newtype wrapper outside Mz. Might want to add the `Arc` implementations, but right now I have nothing blocking on it.

PR #812, opened 2026-07-31
corgi reduce: delta-proportional presentation

frankmcsherry:
The reduce presentation is the corgi backend's one remaining non-delta-proportional path — 95% of an incremental round's profile, ~50% in `collect_present`'s per-retire O(trace) rescan alone (measured: scc steady-state at 5.4x vec per round, vs ~1.1x under open-loop concurrency and 2.2x FASTER than vec on large flat scans).

Opening commit: **seek-vs-scan, decided per retire** when the changed-set and trace sizes are known. Narrow changed sets over seekable (single-leaf) keys gallop via `find_ranges` — no key hashing; broad churn keeps the scan. Steady-state reach improves 4.51s → 2.10s, past the vec backend (0.84x); scc improves 794 → 597ms/round, with the residual attributed to genuinely-broad cascade retires.

Planned follow-ons in this PR's arc: the windowed reduce-presentation seam (the reduce sibling of #809's join reshape), then presentation memoization behind it — the pieces that own scc's remaining gap.

### DIFF
--- a/interactive/src/corgi/reduce.rs
+++ b/interactive/src/corgi/reduce.rs
@@ -37,7 +37,7 @@ use differential_dataflow::trace::chunk::ChunkBatch;
 use differential_dataflow::operators::int_proxy::ProxyBridge;
 use differential_dataflow::operators::int_proxy::reduce::{ProxyReduceBackend, ReduceInstance, ReduceWindow};
 
-use corgi::arrange::{gather, gather_lanes, sort_blocks};
+use corgi::arrange::{find_ranges, gather, gather_lanes, sort_blocks};
 use corgi::{Bounds, Shape, Value as CValue};
 
 use crate::corgi::col_times::ColTime;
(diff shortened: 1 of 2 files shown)


interactive/src/corgi/reduce.rs
```diff
@@ -168,6 +168,19 @@ fn concat_columns(blocks: &[CValue]) -> CValue {
 /// relied upon — so the raw two's-complement `u64` is correct even for negative ints (no swizzle).
 /// Applied CONSISTENTLY at every id site (both value presentations AND the freshly-produced
 /// `reduce_brackets` outputs), else `desired − current` nets across mismatched ids for the same value.
+/// The `changed` set as a needle column in the chunks' own key shape — possible exactly
```
antiguru, line 182: Seems to have landed between the doc comment and the next function?

frankmcsherry: Fixed; thanks!

PR #874, opened 2026-09-11
Adapt to timely #817: least()/retain_least(), inspect_core, Collection::delay

frankmcsherry:
Catches differential-dataflow up to timely #817 (`InputCapability::time()` requires `TotalOrder` and returns `Option`; `retain` is `retain_least`; `for_each_time` is `for_each_stamp`; inspect is `inspect_core` + `inspect`). Stacked on #855, whose `retain_stamp` adaptations this builds on; base is `dd-dynamic-stamps` so the diff is only these changes.

- `Collection::inspect_batch` hands the closure the batch's `Stamp<T>` rather than a time, and `Collection::inspect_container` is `inspect_core`, both without bounds.
- `Collection::delay` no longer uses timely's `delay_batch` (bounded, then removed). DD keeps the time in the data, so it delays each stamp element and each update's time through `func` and sends at once under the delayed `CapabilitySet`. Works for partial orders; the spurious `T: Hash` bound is gone.
- capture's CDC progress operator holds a `CapabilitySet` from `retain_stamp`; upsert takes `retain_least` if present; the diagnostics loggers take `least()` if present, dropping a message with no capabilities.
- dogsdogsdogs `half_join` inserts each element of `retain_stamp`.
- Tests: the dynamic-stamps test asserts a singleton stamp explicitly; the wcoj partial-order test uses `inspect`.

Workspace tests pass against timely master at 11beaae3 (the `Cargo.lock` is not tracked, so the git dependency floats there on its own).

https://claude.ai/code/session_012k2GSwxmvD2LvckkoXi6GK

### DIFF
--- a/diagnostics/src/logging.rs
+++ b/diagnostics/src/logging.rs
@@ -448,46 +448,46 @@ fn construct_timely<'scope>(
                 let mut chs = ch_act.session(&cap);
                 let mut els = el_act.session(&cap);
                 let mut msgs = msg_act.session(&cap);
-                let ts = *cap.time();
-
-                for (event_time, event) in data.drain(..) {
-                    match event {
-                        TimelyEvent::Operates(e) => {
-                            ops.give(((e.id, e.name.clone(), e.addr.clone()), ts, 1i64));
-                            state.operators.insert(e.id, e);
-                        }
-                        TimelyEvent::Shutdown(e) => {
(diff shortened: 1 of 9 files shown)


antiguru (approved):
Looks good!

frankmcsherry:
Folded into #855 (branch `dd-dynamic-stamps`), along with fix-ups for timely #818.

PR #877, opened 2026-09-14
Adapt to timely #817, #818, #819

frankmcsherry:
Catches differential-dataflow up to timely master (4eb3ee09): `InputCapability::time()` requires `TotalOrder` and returns `Option` (#817), operators that read a time off a capability are removed (#818), and `unary_notify` is removed (#819).
`master-next` does not build against timely master without this.

- The three `leave_dynamic` operators (vec, columnar, corgi) hold a capability per element of the message's stamp, truncated as the records are. Their time is a partial order, so there is no least element to take.
- `Collection::delay` no longer uses timely's `delay_batch`. It delays each stamp element and each update's time through `func` and sends at once under the delayed capabilities, which works for partial orders; the `T: Hash` bound is gone.
- `Collection::inspect_batch` hands the closure the batch's `Stamp<T>`; `Collection::inspect_container` is renamed `inspect_core`.
- upsert takes `retain_least`; capture's CDC progress operator holds a `CapabilitySet` from `retain_stamp`; dogsdogsdogs `half_join` inserts each element of `retain_stamp`.
- The diagnostics loggers take `least()` and skip a message with an empty stamp.
- The dogsdogsdogs examples count with DD's `count`; `graphs-static` no longer delays records to their times, as the reduce acts on record times.

This replaces #874 and the timely parts of #855; #855's stamp routing at the batch-to-records seams is not included.

Tests (all targets), doctests, clippy, and the `explain` release test pass locally against timely 4eb3ee09.

https://claude.ai/code/session_01SFY2AqDJTLX9TRGRuGR8dQ


antiguru (approved):
Looks good, thanks!

